PR #7, opened 2025-01-08
Implementing minimum API for clients

WendelHime:


hwh33 (approved):
LGTM!

> Well, unless we want to wait for the VPN stuff to be done so we can merge them at the same time.

I say we go ahead and merge

PR #14, opened 2025-02-03
adding proxyless transport

WendelHime:
This pull request add a proxyless transport with a feature flag for enabling it. The feature can be executed by running the radiance app and providing the flag values following [packet manipulation examples](https://pkg.go.dev/github.com/Jigsaw-Code/outline-sdk/x/configurl#hdr-Packet_manipulation):
```bash
go run cmd/main.go -addr "localhost:8080" -proxyless-config "disorder:0|split:123"
```

You should be able to send a request with:
```bash
curl -v -x localhost:8080 -L https://ipconfig.io
```

If the proxyless dialer was able to successfully send the request, there won't be traffic to the proxy (until it fails or all other conditions are unmet) and that domain will likely use proxyless for the next requests.


WendelHime:
I'm looking at this again and just found a issue, moving to draft

transport/proxyless/stream_dialer.go
```diff
@@ -0,0 +1,126 @@
...
+	NumberOfTries int
+	LastResult    bool
+	LastSuccess   int64
+	ConfigText    string
+}
+
+type contextKey string
+
+// AddrContextKey is the context key used to store the address of the target server.
+var AddrContextKey contextKey = "addr"
```
hwh33, line 33: This shouldn't be exported as it is not used outside of this package.

transport/proxyless/stream_dialer.go
```diff
@@ -0,0 +1,126 @@
...
+	d.upstreamStatusCache[remoteAddr] = status
+}
+
+// DialStream tries to connect to the upstream by using the proxyless configuration if the conditions are met.
+func (d *StreamDialer) DialStream(ctx context.Context, remoteAddr string) (transport.StreamConn, error) {
+	status := d.getUpstreamStatus(remoteAddr)
+	if status.haveNeverTriedProxyless() ||
+		status.itWorkedOnLastTry() ||
+		status.haveNewConfig(d.currentConfig) ||
+		status.lastTryWasLongAgo() {
```
hwh33, line 96: Nice, I love how readable this is!

proxy.go
```diff
@@ -55,25 +50,55 @@ func (h *proxyHandler) handleConnect(proxyResp http.ResponseWriter, proxyReq *ht
...
-	)
-	connectReq.Header.Set(authTokenHeader, h.authToken)
-	if err = connectReq.Write(targetConn); err != nil {
-		sendError(proxyResp, "Failed to write connect request to proxy", http.StatusInternalServerError, err)
-		return
+	var targetConn transport.StreamConn
+	var proxylessFailedOrSkipped bool
+
+	if h.proxylessDialer != nil {
+		targetConn, err = h.proxylessDialer.DialStream(proxyReq.Context(), proxyReq.Host)
```
hwh33, line 63: We need a timeout here or we'll never try to dial through the proxy when proxyless dials hang.  I think 10 seconds should be okay.  Sounds like technically a TCP connection could successfully establish even with 21 seconds of latency and retries, but I can't imagine that's common: https://forum.netgate.com/topic/85793/tcp-timeouts.

hwh33 (commented):
Nice, this is looking really good.  I was reviewing this when you pushed the most recent changes, so I think a couple of my comments may be outdated at this point.

The only big thing missing here is persistence of proxyless dial results (persistence meaning they should be saved to disk, so they persist even when the application is restarted).

hwh33:
> The only big thing missing here is persistence of proxyless dial results (persistence meaning they should be saved to disk, so they persist even when the application is restarted).

@WendelHime and I just discussed this on Slack.  I was under the impression that we were using local storage in Radiance already (to store fetched config), but I see now that I was mistaken.  Let's just punt on persisting proxyless dial results for now.  We can add this when we start using local storage for things like the config.  @WendelHime could you just add a TODO somewhere in the code noting that we want to persist proxyless dial results?

radiance.go
```diff
@@ -61,8 +63,9 @@ type ProxyStatus struct {
 // Radiance is a local server that proxies all requests to a remote proxy server over a transport.StreamDialer.
 // TODO: tunStatus need to be updated when TUN is active
 type Radiance struct {
-	srv         httpServer
-	confHandler configHandler
+	srv                  httpServer
+	confHandler          configHandler
+	activeProxyConfIndex *atomic.Int64
```
hwh33, line 68: This is only used to fetch the proxy location, so we could just save the proxy location right?

radiance.go
```diff
@@ -88,32 +92,49 @@ func NewRadiance() *Radiance {
...
 
-	dialer, err := transport.DialerFrom(conf)
+	var proxyConf, proxylessConf *config.Config
+	for i, conf := range configs {
+		if conf.GetConnectCfgProxyless() != nil {
+			proxylessConf = conf
+		}
+		proxyConf = conf
+		r.activeProxyConfIndex.Store(int64(i))
+	}
```
hwh33, line 109: Currently Radiance [grabs the first proxy config](https://github.com/getlantern/radiance/blob/019f85db047e165fc13df5a2e6506cdc1d73d321/config/config.go#L72).  This will instead grab the last.  I'm not sure how much that matters, just pointing this out.

cc @garmr-ulfr

proxy.go
```diff
@@ -55,39 +51,79 @@ func (h *proxyHandler) handleConnect(proxyResp http.ResponseWriter, proxyReq *ht
...
+
+	return <-errCh
+}
+
+func copyData(dst io.Writer, src io.Reader, errCh chan error) {
+	_, err := io.Copy(dst, src)
+	errCh <- err
+}
+
+func (h *proxyHandler) tryProxylessConnect(r *http.Request, clientConn net.Conn) error {
```
hwh33, line 110: Nice, this is cleaner in this helper function

proxy.go
```diff
@@ -55,39 +51,79 @@ func (h *proxyHandler) handleConnect(proxyResp http.ResponseWriter, proxyReq *ht
...
+		sendError(proxyResp, "Error creating connect request", http.StatusInternalServerError, err)
+		return
+	}
 	connectReq.Header.Set(authTokenHeader, h.authToken)
+
 	if err = connectReq.Write(targetConn); err != nil {
 		sendError(proxyResp, "Failed to write connect request to proxy", http.StatusInternalServerError, err)
 		return
 	}
+	h.pipeData(targetConn, clientConn)
```
hwh33, line 92: This error is ignored -- I think it should be logged if it's not an EOF.

radiance.go
garmr-ulfr: It shouldn't be a problem. Radiance only grabs the first one because it only keeps one config. Lantern-cloud only returns one config to radiance currently anyway.

proxy.go
```diff
@@ -55,39 +51,81 @@ func (h *proxyHandler) handleConnect(proxyResp http.ResponseWriter, proxyReq *ht
...
+		sendError(proxyResp, "Error creating connect request", http.StatusInternalServerError, err)
+		return
+	}
 	connectReq.Header.Set(authTokenHeader, h.authToken)
+
 	if err = connectReq.Write(targetConn); err != nil {
 		sendError(proxyResp, "Failed to write connect request to proxy", http.StatusInternalServerError, err)
 		return
 	}
+	if err = h.pipeData(targetConn, clientConn); err != nil && err != io.EOF {
```
hwh33, line 92: Nit: `err != io.EOF` => `!errors.Is(err, io.EOF)`

hwh33 (approved):
One small nit, but otherwise this looks good!  Nice work @WendelHime 🚀

PR #15, opened 2025-02-13
port algeneva protocol

garmr-ulfr:
This adds support for algeneva.


garmr-ulfr:
I still need to update the README

transport/algeneva/stream_dialer.go
```diff
@@ -0,0 +1,102 @@
...
+Package algeneva provides a [transport.StreamDialer] that uses the application layer geneva protocol to
+route traffic through a proxy server.
+
+HTTP strategies can be found here:
+https://github.com/getlantern/algeneva/blob/main/strategies.go
+
+Application layer geneva modifies the request-line or headers of the request using the specified strategy.
+Any further modifications to these fields essentially breaks the geneva protocol. Due to this, it
+is recommended to only wrap a StreamDialer that does not modify the request-line or headers for it
+to be effective, such as the [transport.TCPDialer].
```
hwh33, line 11: It shouldn't matter, right?  algeneva is agnostic to the payload it carries and wrapping algeneva in anything visible at the application layer defeats the purpose of algeneva, right?

I'm asking because I think this comment can only cause confusion.  The statement essentially applies to every transport.  Does that make sense?

transport/algeneva/stream_dialer_test.go
```diff
@@ -0,0 +1,126 @@
...
+	args := m.Called()
+	return args.Error(0)
+}
+
+func (m *mockStreamConn) CloseWrite() error {
+	args := m.Called()
+	return args.Error(0)
+}
+
+// TODO: fix this test
```
hwh33, line 98: ?

hwh33 (commented):
Mostly looks good!  Just a couple of small questions / comments

radiance.go
```diff
@@ -160,7 +162,7 @@ func (r *Radiance) setStatus(connected bool, status TUNStatus) {
 	r.statusMutex.Unlock()
 
 	// send notifications in a separate goroutine to avoid blocking the Radiance main loop
-	go r.notifyListeners(connected)
+	// go r.notifyListeners(connected)
```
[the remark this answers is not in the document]
garmr-ulfr: Oops.. I did that just for testing against http_proxy. I haven't added algeneva to the assignable proxies for radiance yet so I had to use a local config. It was complaining when I modified `ConfigHandler` and I forgot to un-comment it.

transport/algeneva/stream_dialer.go
garmr-ulfr: > It shouldn't matter, right? algeneva is agnostic to the payload it carries

To the payload, yes, but not the inner dialer. This is (or should be) saying that wrapping a stream dialer _with_ algeneva (`algeneva(innerDialer)`) could render it ineffective. _e.g_ If a TLS dialer is passed as `innerSD`, `algeneva.NewStreamDialer(tlsStreamDialer, cfg)`, then when algeneva sends the modified request `innerSD` will encrypt it and the algeneva modifications won't matter.

transport/algeneva/stream_dialer_test.go
garmr-ulfr: That's the test I reference in my comment above. I need to create an algeneva listener or add a mock listener to handle the websocket handshake so I can test that calling `CloseRead` and `CloseWrite` on the returned conn will propagate to the inner conn.

hwh33: Oh sorry, I missed that part of your comment!

transport/algeneva/stream_dialer.go
garmr-ulfr: Discussed on a call. It's been removed as it's confusing.

garmr-ulfr:
Test fixed, and README updated.

transport/algeneva/stream_dialer_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}{
+		{
+			name: "success",
+			cfg: &config.Config{
+				Addr:           "1.1.1.1",
+				CertPem:        certPem,
+				ProtocolConfig: protocolConfig,
+			},
+			assert: func(t *testing.T, got transport.StreamDialer, cfg *config.Config, err error) {
+
```
WendelHime, line 44:
```suggestion
				assert.NoError(t, err)
				assert.NotNil(t, got)
```

We could also assert if the config opts are the same as the config expects.

transport/algeneva/stream_dialer_test.go
```diff
@@ -0,0 +1,129 @@
...
+MIIBhTCCASugAwIBAgIQIRi6zePL6mKjOipn+dNuaTAKBggqhkjOPQQDAjASMRAw
+DgYDVQQKEwdBY21lIENvMB4XDTE3MTAyMDE5NDMwNloXDTE4MTAyMDE5NDMwNlow
+EjEQMA4GA1UEChMHQWNtZSBDbzBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABD0d
+7VNhbWvZLWPuj/RtHFjvtJBEwOkhbN/BnnE8rnZR8+sbwnc/KhCk3FhnpHZnQz7B
+5aETbbIgmuvewdjvSBSjYzBhMA4GA1UdDwEB/wQEAwICpDATBgNVHSUEDDAKBggr
+BgEFBQcDATAPBgNVHRMBAf8EBTADAQH/MCkGA1UdEQQiMCCCDmxvY2FsaG9zdDo1
+NDUzgg4xMjcuMC4wLjE6NTQ1MzAKBggqhkjOPQQDAgNIADBFAiEA2zpJEPQyz6/l
+Wf86aX6PepsntZv2GYlA5UpabfT2EZICICpJ5h/iI+i341gBmLiAFQOyTDT+/wQc
+6MF9+Yw1Yy0t
+-----END CERTIFICATE-----`)
```
WendelHime, line 27: (optional) If by any reason in the future we need to use a certPem, we could add a testhelper for generating those certs

```go
func generateCertPEM(t *testing.T) []byte {
	// generating private key
	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
	require.NoError(t, err)

	// certificate template
	serialNumber, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
	require.NoError(t, err)
	template := x509.Certificate{
		SerialNumber: serialNumber,
		Subject: pkix.Name{
			Organization: []string{"org test"},
		},
		NotBefore:             time.Now(),
		NotAfter:              time.Now().Add(1 * time.Hour),
		KeyUsage:              x509.KeyUsageEncipherOnly | x509.KeyUsageDigitalSignature,
		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
		BasicConstraintsValid: true,
	}

	// self-signing
	certDER, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
	require.NoError(t, err)

	var certPEM bytes.Buffer
	require.NoError(t, pem.Encode(&certPEM, &pem.Block{Type: "CERTIFICATE", Bytes: certDER}))

	return certPEM.Bytes()
}
```

WendelHime: I guess the positive side of using generating a certificate during the test is it won't expire and we can modify for other tests

WendelHime (changes requested):
This is awesome 🚀 !
I've synced with main for avoiding conflicts (hoping you don't mind). 

There's one small suggestion for assertion and the other comment is totally optional

WendelHime (approved):

hwh33 (approved):

PR #16, opened 2025-02-20
Add crash and error reporting to sentry

myleshorton:
This closes https://github.com/getlantern/engineering/issues/1925 and uses sentry to report panics and errors.


common/reporting/sentry_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/getsentry/sentry-go"
+)
+
+func TestInitSentry(t *testing.T) {
+	// Call the function to test
+	Init()
+
+	defer sentry.Flush(2 * time.Second)
+	sentry.CaptureException(errors.New("Hello, Sentry!"))
+}
```
WendelHime, line 17: Since the main goal here is to test the Sentry initialization by providing hard-coded options (and make sure it doesn't panic), we could use `assert.NotPanics` (from `"github.com/stretchr/testify/assert"`) and remove the sentry calls since we're not validating if sentry is working or not. If we want to test the sentry integration, we need to update the HTTP transport for mocking the responses and avoid sending the error to them.
```go
	assert.NotPanics(t, func() { Init() })
```

WendelHime: We could also add a test for `PanicListener` and use the `BeforeSend` option to verify if the event level was defined as `Fatal`

WendelHime (approved):
Added a few suggestions and pushed a commit for fixing my broken tests (sorry!)

common/reporting/sentry_test.go
myleshorton: Cool yeah tbh the intent was actually to test if it's working or not, but I'll just remove it for now since it is working.

PR #17, opened 2025-02-24
Adding proxyless outbound

WendelHime:
This pull request focus on adding a proxyless outbound that should use outline-sdk configurl techniques (such as split, disorder and tlsfrag) created at the proxyless transport (from #14). There's also a few modifications to the proxyless transport because it wasn't receiving a inner stream dialer and we need to send the traffic through an outbox stream dialer so we can use the created tun for sending and receiving data.


transport/proxyless/stream_dialer.go
```diff
@@ -107,25 +113,30 @@ func (d *StreamDialer) updateUpstreamStatus(remoteAddr, configText string, succe
 }
 
 // DialStream tries to connect to the upstream by using the proxyless configuration if the conditions are met.
-func (d *StreamDialer) DialStream(ctx context.Context, remoteAddr string) (transport.StreamConn, error) {
-	status := d.getUpstreamStatus(remoteAddr)
+// When tunneling, please provide a context with RemoteAddrContextKey with the value as ip:port format so
+// the dialer use the IP for dialing instead of using the domain
+func (d *StreamDialer) DialStream(ctx context.Context, domain string) (transport.StreamConn, error) {
```
garmr-ulfr, line 118: This should probably be the other way around: store domain in ctx and pass `remoteAddr` to `DialStream`, so that it conforms to the other transports.

WendelHime: So in the case of proxy, I'd resolve the domain first, add the domain to the context and dial with the ip port?

garmr-ulfr: Like running radiance as a proxy? The caller would add the domain to the context and then pass the context and ip:port to DialStream. 
One thing you could do is, if the domain doesn't exist in the context then use `net.LookupAddr` to do a reverse lookup for the domain.

WendelHime: Got it, thanks for the suggestions and the help! Will add it tomorrow!

WendelHime: For the VPN we already have the sniff so that's cool, for the proxy part I'll need to lookup for the IP (for the proxy we won't be tunneling as far as I understood)

outbounds/proxyless/proxyless.go
```diff
@@ -0,0 +1,104 @@
...
+// for dialing
+func NewProxylessOutbound(ctx context.Context, router adapter.Router, logger log.ContextLogger, tag string, options option.DialerOptions) (adapter.Outbound, error) {
+	glog.Debug("creating outbound dialer")
+	outboundDialer, err := dialer.New(ctx, options)
+	if err != nil {
+		return nil, err
+	}
+
+	glog.Debug("getting config")
+	ch := config.NewConfigHandler(time.Minute * 10)
```
hwh33, line 44: This will kick off an independent config polling routine.  We don't want that.  We want the client to manage config fetching centrally and respond to config updates as needed.  A config update to proxyless dialing for example would involve re-loading the proxyless dialer (calling `NewProxylessOutbound` again I think).

You should be able to consult how flashlight does this as an example.

cc @garmr-ulfr since it looks like this copies code from a commit of yours in algeneva_outbound.go

outbounds/proxyless/proxyless.go
```diff
@@ -0,0 +1,104 @@
...
+// wrapper around sing-box's network.Dialer to implement streamDialer interface to pass to a
+// stream dialer as innerSD
+type sboxSD struct {
+	outSD  network.Dialer
+	logger log.ContextLogger
+}
+
+func (s *sboxSD) DialStream(ctx context.Context, addr string) (otransport.StreamConn, error) {
+	s.logger.InfoContext(ctx, "proxyless sboxSD dialing ", addr)
+	destination := metadata.ParseSocksaddr(addr)
```
hwh33, line 97: Just curious -- why is the input a SOCKS address?

client/rules.go
```diff
@@ -14,21 +16,15 @@ func getopts() option.Options {
...
-		// 	},
+		DNS: &option.DNSOptions{
+			Servers: []option.DNSServerOptions{
+				{
+					Tag:      "google",
+					Address:  "8.8.8.8",
+					Strategy: option.DomainStrategy(dns.DomainStrategyUseIPv4),
+				},
+			},
+		},
```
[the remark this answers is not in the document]
WendelHime: It was required in my machine (linux pop-os, at mac it worked smoothly without the DNS config), since the DNS was unreachable when starting tunneling we had to add a DNS server option

outbounds/proxyless/proxyless.go
WendelHime: The `network.Dialer` interface from `github.com/sagernet/sing/common/network` expects a SocksAddr (in the same way as the outbound does) but I don't know if there's an sing-box special reason

proxy.go
```diff
@@ -129,6 +137,19 @@ func (h *proxyHandler) tryProxylessConnect(r *http.Request, clientConn net.Conn)
 	return h.pipeData(targetConn, clientConn)
 }
 
+func raddrFromRequest(r *http.Request) (string, error) {
+	ips, err := net.LookupIP(r.Host)
+	if err != nil {
+		return "", log.Errorf("failed to lookup for host IPs: %w", err)
+	}
```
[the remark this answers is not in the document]
WendelHime: I can try to provide a domain and test if it works but I believe the outbound inner dialer would expect an IP instead of the domain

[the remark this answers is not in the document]
WendelHime: No it works with domain or ip:port, I'll change it to just provide the domain

transport/proxyless/disorder.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+	"strconv"
+
+	"github.com/Jigsaw-Code/outline-sdk/transport"
+	"github.com/Jigsaw-Code/outline-sdk/x/configurl"
+	"github.com/Jigsaw-Code/outline-sdk/x/disorder"
+)
+
+func registerDisorderDialer(r configurl.TypeRegistry[transport.StreamDialer], typeID string, newSD configurl.BuildFunc[transport.StreamDialer]) {
+	r.RegisterType(typeID, func(ctx context.Context, config *configurl.Config) (transport.StreamDialer, error) {
```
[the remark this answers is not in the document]
WendelHime: Yes, after discussing with Patrick we've noticed that we didn't want to use `configurl` because it didn't support providing a inner stream dialer and in our case it's crucial for our implementations to use the outbound inner stream dialer since we make the requests through the tunneling. After we noticed this issue, I've noticed that we could customize those configurl container provider for using a custom stream dialer and updated the code for registering the same proxyless dialers that we want to support at this moment (disorder, split and tlsfrag) and update the provider implementation so it uses our outbound stream dialer as a inner SD.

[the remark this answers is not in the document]
WendelHime: So those register functions are basically copy pasted from outline sdk configurl (they're private functions I had to do that) and they're used here: https://github.com/getlantern/radiance/pull/17/files#diff-5286a6a77c4add7ee502e9704c32d1cc894250ae87424db230586aef57603baeR69-R82

proxy.go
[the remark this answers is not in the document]
WendelHime: Fixed!

client/rules.go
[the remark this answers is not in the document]
WendelHime: BTW that was happening because `curl` tries to resolve the domain on request by reaching a DNS and 127.0.0.1:53 wasn't available while tunneling

outbounds/proxyless/proxyless.go
WendelHime: I've updated the code for receiving the config from outbound options, now we need to update `getopts` for building outbounds based on config

garmr-ulfr: > cc @garmr-ulfr since it looks like this copies code from a commit of yours in algeneva_outbound.go
Yes. Again, I want to stress that what I did was just to get it working without concern for correctness, organization, etc. All of it will be cleaned up and slightly refactored.

WendelHime:
I've just noticed this isn't ready for review because we're not using an alternative if proxyless fail to dial (in the same way we do at `proxy.go`) so I'm returning this to draft

transport/proxyless/stream_dialer.go
garmr-ulfr: >  ...if the machine doesn't have a DNS server running locally or it's unreachable due to tunneling...

Just FYI, the DNS rules we added tells sing-box to resolve the domains itself instead of handing it off to the host. We'll be using those (or something similar) for the VPN.

client/rules.go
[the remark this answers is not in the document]
garmr-ulfr: @hwh33, we don't _have_ to specify DNS options, but VPNs usually handle DNS queries themselves to prevent ISPs from monitoring them. They will use their own DNS server, if they have them (NordVPN, ProtonVPN), if not, they use Google, 8.8.8.8, or Cloudflare, 1.1.1.1. (OpenVpn, ExpressVPN). I think we'll want to do the same.

WendelHime:
> I've just noticed this isn't ready for review because we're not using an alternative if proxyless fail to dial (in the same way we do at `proxy.go`) so I'm returning this to draft

I've removed the fallback dialer support because I think this should be responsibility of another outbound that should deal with other outbounds.

This is ready for review again!

client/rules.go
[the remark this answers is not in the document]
myleshorton: We probably want to unify our DNS providers. The problem with Google is that it's likely blocked, and the problem with running our own is that it likely will get blocked.

We probably want to have a canonical list that everything uses, but here's what we're using in the smart dialer in kindling -- basically cloudflare, alibaba, and opendns:

https://github.com/getlantern/kindling/blob/main/smart_dialer_config.yml

myleshorton:
> I've just noticed this isn't ready for review because we're not using an alternative if proxyless fail to dial (in the same way we do at `proxy.go`) so I'm returning this to draft

I think using an alternative if proxyless fails is a *much* bigger lift that should likely be handled as a completely separate protocol -- i.e. a dialer along the lines of what we do now in flashlight that uses all the proxies we know about, with proxyless just being one. I don't think that that should a part of this PR or a part of proxyless at all.

myleshorton:
To me, the other issue here is that this should really be a protocol *in sing-box* not in radiance.

myleshorton:
Like it should be more like https://github.com/getlantern/sing-box/tree/amnezia, although as an entirely new protocol.

garmr-ulfr:
@myleshorton, sing-box already has a "proxyless" protocol (in the most literal definition), `direct`. 

> I think using an alternative if proxyless fails is a much bigger lift that should likely be handled as a completely separate protocol
> ...
> although as an entirely new protocol.

I think we would just need to expand on the [selector outbound](https://sing-box.sagernet.org/configuration/outbound/selector/) by adding an option for "mode" (manual or auto). In auto mode, we could use the [URL test outbound](https://sing-box.sagernet.org/configuration/outbound/urltest/) to determine which outbound to use.

myleshorton:
Right, but we want to add the outline smart dialer - that's the work to be done here.

garmr-ulfr:
> we want to add the outline smart dialer 

Oh, that's right. Never mind.

myleshorton:
Yeah so unfortunately I think this should mostly be about adding the outline smart dialer/proxyless dialing to sing-box a la https://github.com/getlantern/sing-box/tree/amnezia.

I think radiance is then more about:

1. Fetching sing-box compatible configs
2. Providing an interface/API to the frontend
3. Wrapping sing-box to some relatively minimal extent (providing the TUN config, fetching the ancillary configs like the routing rules, etc).

myleshorton:
> I think we would just need to expand on the [selector outbound](https://sing-box.sagernet.org/configuration/outbound/selector/) by adding an option for "mode" (manual or auto). In auto mode, we could use the [URL test outbound](https://sing-box.sagernet.org/configuration/outbound/urltest/) to determine which outbound to use.

This sounds promising to me. I just think that work shouldn't be a part of the proxyless stuff -- should be a separate protocols or outbound or whatever the operates at a higher level where proxyless is one of the available dialers. That should probably look very similar to the current flashlight dialing.

hwh33:
I think this has effectively been superseded by https://github.com/getlantern/radiance/pull/22.  Is that right, @WendelHime?

PR #18, opened 2025-02-24
Add skeleton API

hwh33:
This PR just adds a skeleton of the API defined in https://github.com/getlantern/engineering/issues/1900.  I also removed some of (but not all of) the old API which will no longer be needed.  User management is not included yet as that is still in design.

### DIFF
--- a/radiance.go
+++ b/radiance.go
@@ -7,6 +7,7 @@ package radiance
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -14,8 +15,9 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/getlantern/golog"
 	"github.com/getsentry/sentry-go"
+
+	"github.com/getlantern/golog"
 
 	"github.com/getlantern/radiance/common/reporting"
 	"github.com/getlantern/radiance/config"
@@ -28,6 +30,10 @@ var (
 
 	configPollInterval = 10 * time.Minute
 )
+
+// ErrNotImplemented is returned by functions which have not yet been implemented. The existence of
+// this error is temporary; this will go away when the API stabilized.
+var ErrNotImplemented = errors.New("not yet implemented")
 
 //go:generate mockgen -destination=radiance_mock_test.go -package=radiance github.com/getlantern/radiance httpServer,configHandler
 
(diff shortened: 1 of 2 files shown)


radiance.go
```diff
@@ -201,34 +196,34 @@ func (r *Radiance) setStatus(connected bool, status TUNStatus) {
...
-		select {
-		case listener <- status:
-		default:
-		}
-	}
+type Location struct {
+	Country   string
+	City      string
+	Lat, Long float32
 }
```
garmr-ulfr, line 206: A location type already [exists](https://github.com/getlantern/radiance/blob/main/config/types.proto#L47) as part of `config.Config`, but as `ProxyLocation`. We could define an alias for it if we want it to be named "Location", but I think "ProxyLocation" is more descriptive.

radiance.go
```diff
@@ -175,21 +154,37 @@ func (r *Radiance) Shutdown(ctx context.Context) error {
...
 	return nil
 }
 
+// StartVPN starts the local VPN device, configuring routing rules such that network traffic on this
+// machine is sent through this instance of Radiance.
+//
+// This function will be implemented as part of https://github.com/getlantern/engineering/issues/1883
+func (r *Radiance) StartVPN() error {
+	// TODO: implement me!
+	return nil
```
garmr-ulfr, line 168: IMO, we should return `ErrNotImplemented` instead of `nil` for functions that are defined but not implement (yet). That way, we'll be less likely to miss it since it will always fail tests and show up in logs. That being said, it is fine as it is if you prefer to leave it.

radiance.go
```diff
@@ -242,13 +237,17 @@ func (r *Radiance) ActiveProxyLocation(ctx context.Context) string {
...
+type IssueReport struct {
+}
+
+// ReportIssue submits an issue report to the back-end.
+//
+// This function will be implemented as part of https://github.com/getlantern/engineering/issues/1921
+func (r *Radiance) ReportIssue(report IssueReport) error {
+	// TODO: implement me!
+	return nil
 }
```
garmr-ulfr, line 253: We don't need to worry about this now, but we should probably move issue  reporting into it's own package.

radiance.go
hwh33: > A location type already [exists](https://github.com/getlantern/radiance/blob/main/config/types.proto#L47) as part of config.Config

We need a type defined here, for use by consumers of this API.  The goal is that someone could use Radiance exclusively via the concise top-level `getlantern/radiance` package.  This is in contrast to `flashlight`, where we ended up with a scattering of API across many packages.

> We could define an alias for it if we want it to be named "Location", but I think "ProxyLocation" is more descriptive.

A type alias makes sense to me I think 🤔.  As far as the name, there's nothing specific here to this being a proxy location.  It is used eventually to represent proxy location, but in its definition here, it's abstract from that use case.

radiance.go
hwh33: Oh, great idea!

radiance.go
hwh33: The logic can live anywhere, but we need this function to live here as part of this API.  See [my comment above](https://github.com/getlantern/radiance/pull/18#discussion_r1970454418) and the [broader work](https://github.com/getlantern/engineering/issues/1900).

radiance.go
garmr-ulfr: > It is used eventually to represent proxy location, but in its definition here, it's abstract from that use case.
Ohh, OK. That makes sense.

Well, we could type alias it and just grab it from the config or leave it as is and just assign the values directly. There's only 4 so it's not that big of a deal.

garmr-ulfr (approved):
LGTM

radiance.go
hwh33: Oh, to be clear, the bit you quoted above is in reference to the name (`Location` vs `ProxyLocation`)

hwh33: That said, I just realized that if we're type aliasing it to something called `ProxyLocation`, it's a little awkward to pretend it's abstracted lol

hwh33:
Thanks @garmr-ulfr!

PR #21, opened 2025-03-03
Vpn client

garmr-ulfr:
This is adds a rudimentary VPN client. The client only creates a TUN device and configures routing rules so that all traffic is routed through the TUN device.


client/algeneva/inbound.go
WendelHime, line 1: The algeneva related stuff is going to be moved to `github.com/getlantern/sing-box` right? If yes, we can probably remove this from the PR since it's out of scope. Related files:

- `client/algeneva/inbound.go`
- `client/algeneva/options.go`
- `client/algeneva/outbound.go`

WendelHime (approved):
Added a small suggestion, everything else LGTM! I'm able to run and it's tunneling traffic as expected

client/algeneva/inbound.go
garmr-ulfr: Oh yeah. Thanks

PR #22, opened 2025-03-05
Adding outline outbound

WendelHime:
This pull request add outline outbound and rules sending traffic through the outline outbound. There's also some test fixes and I've replaced sing-box at go.mod for referring to `getlantern/sing-box` for now.

You can manual test this by building, running lantern and sending a request to `google.com` or `api.iantem.io`


client/boxoptions/options.go
```diff
@@ -19,32 +20,38 @@ var boxOptions = option.Options{
...
+			},
+			{
+				Tag:             "dns-cloudflare-doh",
+				Address:         "https://cloudflare-dns.com/dns-query",
+				AddressResolver: "dns-cloudflare-dot",
+			},
+			{
+				Tag:             "dns-sb-doh",
+				Address:         "https://doh.dns.sb/dns-query",
+				AddressResolver: "dns-sb-dot",
```
myleshorton, line 47: I think we should probably use the same DNS we us here:

https://github.com/getlantern/kindling/blob/main/smart_dialer_config.yml

myleshorton (approved):
i think this generally looks good for now. i think we ultimately want the list of domains to come from our server/remote routing rules, but I think this works for now.

client/boxoptions/options.go
```diff
@@ -77,57 +84,109 @@ var boxOptions = option.Options{
...
+						HTTPS: &option.HTTPSEntryConfig{
+							Name:    "doh.dns.sb",
+							Address: "https://doh.dns.sb/dns-query",
+						},
+					},
+					{
+						System: &struct{}{},
+					},
+				},
+				TLS:         []string{"", "split:1", "split:2,20*5", "split:200|disorder:1", "tlsfrag:1"},
```
hwh33, line 119: `split` and `disorder` are mutations at the TCP level, not the TLS level.  But it seems like those things are now in our sing-box fork, so maybe that's slightly out of scope of this PR.  Just noting that this is a little odd 🤷

radiance_test.go
```diff
@@ -132,7 +132,7 @@ func TestNewRadiance(t *testing.T) {
 		r, err := NewRadiance()
 		assert.NoError(t, err)
 		assert.NotNil(t, r)
-		assert.NotNil(t, r.confHandler)
+		//assert.NotNil(t, r.confHandler)
```
hwh33, line 135: Is this supposed to remain commented out like this?  If so, could you provide a comment explaining why?  Alternatively, if this part of the test doesn't make sense anymore, we should just remove it.

hwh33 (approved):
A couple of small things, but mostly looks good

client/boxoptions/options.go
WendelHime: This is how the stream dialer receive those techniques from the config

WendelHime: The [TLS name](https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/smart/stream_dialer.go#L99) in this case is just how they receive the techniques and it's not exclusive to TLS

radiance_test.go
WendelHime: I'll add a line explaining, but basically, currently we're not fetching the configuration from the backend and not even creating the config handler when starting the application, so that was breaking the tests

client/boxoptions/options.go
WendelHime: But I agree it's weird and we could probably update and rename at the sing-box implementation

radiance_test.go
WendelHime: Fixed!

client/boxoptions/options.go
myleshorton: Right -- the idea is that these are the techniques it will use for TLS traffic -- i.e. for https. It doesn't have support for doing anything for arbitrary TCP traffic.

myleshorton: > The [TLS name](https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/smart/stream_dialer.go#L99) in this case is just how they receive the techniques and it's not exclusive to TLS

Pretty sure it is exclusive to TLS -- I don't think the smart dialer does anything with other types of traffic.

WendelHime: > I don't think the smart dialer does anything with other types of traffic.

Yes, after [looking through the code](https://github.com/Jigsaw-Code/outline-sdk/blob/ac9482090891ae27986d7d5e83408f3231471f1e/x/smart/stream_dialer.go#L244-L272) it searches for the TLS dialer that works with those configurations in parallel (by dialing and handshaking) and use the selected dialer with the technique for stream dial

WendelHime: Looking again through the code it calls it a `tlsDialer` but it could be a TCP dialer wrapped by the split implementation returned from the `config.NewStreamDialer`

WendelHime: So it dials to the test domain with the stream dialer that not necessarily is a TLS dialer, wraps the connection as a TLS, do the handshake and the selected dialer is returned

WendelHime: We could probably rename the `TLS` property to proxyless techniques or something else that makes sense because it's not really related to TLS but that would be different then the smart dialer config from outline. Should we still change that?

WendelHime: hmmmm ok now I've debugged locally and whatever dialer being created from the given config, it's using the `transport.HappyEyeballsStreamDialer` as a inner dialer (that uses the outbound dialer created at the sing-box PR)

garmr-ulfr (approved):
LGTM

client/boxoptions/options.go
hwh33: Oh, interesting!  Odd it restricts TCP-level stuff to TLS, but I guess their use case is narrow.

Sorry for the distraction!

PR #23, opened 2025-03-05
Report sing-box version in config requests

myleshorton:


myleshorton:
This does use a much newer protoc...

myleshorton:
The same test is also failing on main

myleshorton:
Pulling this in...

PR #24, opened 2025-03-06
Add user management API

hwh33:
As defined in https://github.com/getlantern/engineering/issues/1916.  This is the API to be implemented by sub-tasks of that epic.  Much of this is adapted from the authentication client [already defined in lantern-client](https://github.com/getlantern/lantern-client/blob/aaa6d6a1147747658a6568d7dfce4f704cf7bb9c/internalsdk/auth/auth.go#L27-L49).  The idea is that we can port that code to implement most of this.  I changed the API a bit to:
- avoid leaking things like the salt through the abstraction (WIP; see comments below)
- manage some of the work for the caller, like storing the salt

### DIFF
--- a/user.go
+++ b/user.go
@@ -1,0 +1,131 @@
+package radiance
+
+import (
+	"context"
+	"time"
+)
+
+// The main output of this file is Radiance.GetUser, which provides a hook into all user account
+// functionality.
+
+// DataCapInfo represents information about the data cap for a user account.
+type DataCapInfo struct {
+	BytesAllotted, BytesRemaining int
+	AllotmentStart, AllotmentEnd  time.Time
+}
+
+// Tier is the level of subscription a user is currently at.
+type Tier int
+
+const (
+	TierFree = 0
+	TierPro  = 1
+)
+
+// Subscription holds information about a user's paid subscription.
+type Subscription struct {
+	Tier    Tier
+	Expires time.Time
+}
+
+// Device is a machine registered to a user account (e.g. an Android phone or a Windows desktop).
+type Device struct {
+	Name     string
+	Platform string
+}
+
+// User represents a user account. This may be a free user, associated only with this device or a
+// paid user with a full account.
+type User struct {
+}
+
+// GetUser returns information about the current user.
+func (r *Radiance) GetUser() (*User, error) {
+	// TODO: implement me!
+	return nil, ErrNotImplemented
+}
+
+// Devices returns a list of devices associated with this user account.
+func (u *User) Devices() []Device {
+	// TODO: implement me!
+	return []Device{}
+}
+
+// Subscription returns the subscription status of this user account.
+func (u *User) Subscription() Subscription {
+	// TODO: implement me!
+	return Subscription{}
+}
+
+// DataCapInfo returns information about this user's data cap. Only valid for free accounts.
+func (u *User) DataCapInfo() (*DataCapInfo, error) {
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// StartRecoveryByEmail initializes the account recovery process for the provided email.
+func (u *User) StartRecoveryByEmail(ctx context.Context, email string) error {
+	// TODO: implement me!
+	return ErrNotImplemented
+}
+
+// CompleteRecoveryByEmail completes account recovery using the code received via email.
+func (u *User) CompleteRecoveryByEmail(ctx context.Context, email, code, salt, verifier string) error {
```
hwh33, line 103: This comes from the existing function [here](https://github.com/getlantern/lantern-client/blob/aaa6d6a1147747658a6568d7dfce4f704cf7bb9c/internalsdk/auth/auth.go#L38), but I'm confused about a couple of things:
- Is the user expected to provide the salt or the application?  If the latter, what happens if we don't have the salt stored anywhere?  What if this is a device the user hasn't signed in on before?  If the former, where do they get the salt?  Does it come in via email during the recovery process?
- Why is there both a code and a verifier?  Do these both get sent to the user's email during the recovery process?

user.go
```diff
@@ -0,0 +1,131 @@
...
+
+// StartChangeEmail initializes a change of the email address associated with this user account.
+func (u *User) StartChangeEmail(ctx context.Context, newEmail string) error {
+	// TODO: implement me!
+	return ErrNotImplemented
+}
+
+// CompleteChangeEmail completes a change of the email address associated with this user account,
+// using the code recieved via email.
+func (u *User) CompleteChangeEmail(ctx context.Context, newEmail, code, verifier string) error {
```
hwh33, line 122: Similar to above, why is there both a code and a verifier? Do these both get sent to the user's email during the email-change process?

user.go
hwh33: cc @reflog because the back-end mechanics might be helpful for my understanding

reflog: @hwh33 please see the original PR https://github.com/getlantern/lantern-cloud/pull/353
it has documentation on the SRP flow and the terms used

in general - 'code' is what user gets in the email
verifier - a large number derived from user/pass

hwh33: Thank you, I will take a look!

hwh33: Okay, from what I can tell using the original PR you linked @reflog, the new salt and new verifier come from the client in this case and the server just [accepts what it is told](https://github.com/getlantern/lantern-cloud/pull/353/files#diff-7c0012be06256dc7ab2ac84220cfce0d9e0323da45f0a90dcc9333288b4ed2a7R62-R77).  This implies to me that the client can simply produce a new salt and verifier out of mid-air if needed.  This all makes sense logically as well since the code, delivered via email, verifies the user's access to the configured email account and therefore the user's identity.

Given this, I have removed the salt and verifier parameters from the API here.  These can and should be managed by Radiance under the hood.

user.go
hwh33: For posterity, see resolution [here](https://github.com/getlantern/radiance/pull/24#discussion_r1987975321).

hwh33:
@atavism please let me know if this API works for you team

Also, I understand that there has been some work on integrating login via social platforms (e.g. Google).  I haven't included that since it is still in progress AFAIK.

PR #31, opened 2025-03-13
Add ReportIssue() API

Jovis7:
Closes https://github.com/getlantern/engineering/issues/1921

Changes:
- Add a protobuf for issue reporting
- pull some required code from flashlight (zip related functions in `util` package)
- `backend` package: add a new function with required issue reporting headers
- `app` package: set `Platform` value
- `config` package: save the "country" returned by the backend for issue reporting
- `issue` package: add a function for radiance to send the issue report
- `radiance`: add `ReportIssue()` API


myleshorton:
Just added this PR to your PR that adds the logs directory and should resolve conflicts:

https://github.com/getlantern/radiance/pull/33

config/config.go
```diff
@@ -121,17 +123,17 @@ func (ch *ConfigHandler) fetchLoop(pollInterval time.Duration) {
...
-func (ch *ConfigHandler) GetConfig(ctx context.Context) ([]*Config, error) {
+func (ch *ConfigHandler) GetConfig(ctx context.Context) ([]*Config, string, error) {
 	_cfgRes, err := ch.config.Get(ctx)
 	if err != nil { // ctx expired
-		return nil, err
+		return nil, "", err
 	}
 	cfgRes := _cfgRes.(configResult)
-	return cfgRes.cfg, cfgRes.err
+	return cfgRes.cfg, cfgRes.country, cfgRes.err
```
WendelHime, line 136: Optional: We could probably create a struct like `ConfigResponse` for storing the specific fields we're using from `configResult`,  but this implementation also works

util/zip.go
WendelHime, line 1: Since this is only being used by issue, we can probably move the zip package to the issue package instead of having a `util` package

util/test_zip_src/extra.file
WendelHime, line 1: We can probably rename `test_zip_src` as `test_data`

WendelHime (approved):
LGTM! Left a few suggestions!

config/config.go
Jovis7: Thank you. I will leave it as it is now until we need more fields in the future.

util/zip.go
Jovis7: I completely agree with you. I've made this change.

util/test_zip_src/extra.file
Jovis7: Great suggestion. I've made the change.

Jovis7:
Thanks for reviewing. I am merging this now. Please let me know if anyone has some issue with this issue reporting PR. I will then fix/modify.

PR #34, opened 2025-03-14
Added support for requesting server list and passing the preference during config request

reflog:
closes https://github.com/getlantern/engineering/issues/1965


WendelHime (approved):
LGTM!

config/config.go
```diff
@@ -69,10 +88,41 @@ func NewConfigHandler(pollInterval time.Duration, httpClient *http.Client, user
 	return ch
 }
 
+func (ch *ConfigHandler) SetPreferredServerLocation(country, city string) {
+	ch.preferredServerLocation.Store(&serverLocation{
+		Country: country,
+		City:    city,
+	})
+}
```
garmr-ulfr, line 96: We probably want a way to clear the preferred server location in case the user wants to unset it.

reflog: good call, I'll add that!

myleshorton:
Ok pulling this in thanks @reflog!

PR #36, opened 2025-03-14
Use empty preferred location by default

myleshorton:
Just nits discovered in code review

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -79,6 +79,9 @@ func NewConfigHandler(pollInterval time.Duration, httpClient *http.Client, user 
 		apiClient:               common.NewWebClient(httpClient),
 		preferredServerLocation: atomic.Value{}, // initially, no preference
 	}
+	// Store an empty preferred location to avoid nil pointer dereference
+	ch.preferredServerLocation.Store(&serverLocation{})
+
 	// if err := ch.loadConfig(); err != nil {
 	// 	log.Errorf("failed to load config: %v", err)
 	// }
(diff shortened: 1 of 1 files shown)


myleshorton:
OK I also added a fetch of the new config after the location is set and moved that file to slog.

config/config.go
```diff
@@ -79,8 +78,11 @@ func NewConfigHandler(pollInterval time.Duration, httpClient *http.Client, user
 		apiClient:               common.NewWebClient(httpClient),
 		preferredServerLocation: atomic.Value{}, // initially, no preference
 	}
+	// Store an empty preferred location to avoid nil pointer dereference
```
reflog, line 81: this is not good. setting the value to empty Preferred Region will break the API call, since it assumes that only 'nil' means No Preference and it will try to search for empty city and country

reflog (changes requested):

config/config.go
myleshorton: Does the previous code also specify an empty server and country? The `preferredServerLocation := &serverLocation{}` below?

```go
locationPtr := ch.preferredServerLocation.Load()
preferredServerLocation := &serverLocation{}
if locationPtr != nil {
    preferredServerLocation = locationPtr.(*serverLocation)
}
```

reflog: @myleshorton hah. no, it's me being an idiot. I'm lecturing you about something I didn't do properly myself 😆 
it was supposed to be
```
var preferredServerLocation *PreferredRegion
```
instead of the empty assignment

PR #39, opened 2025-03-17
Libbox switch

garmr-ulfr:


client/service/service.go
```diff
@@ -0,0 +1,152 @@
...
+	//
+	//	ctx = filemanager.WithDefault(ctx, sWorkingPath, sTempPath, sUserID, sGroupID)
+	// ctx = service.ContextWith(ctx, filemgr)
+
+	ctx, cancel := context.WithCancel(ctx)
+	urlTestHistoryStorage := urltest.NewHistoryStorage()
+	ctx = service.ContextWithPtr(ctx, urlTestHistoryStorage)
+
+	if platIfce != nil {
+		service.MustRegister[platform.Interface](ctx, platIfce)
```
atavism, line 78: I noticed you are registering the raw platform interface from Kotlin/Swift directly here. [libbox.NewService](https://github.com/SagerNet/sing-box/blob/dev-next/experimental/libbox/service.go#L57) wraps the platform.Interface in a platformWrapper, which adds extra logic for handling things like [opening the tun device](https://github.com/SagerNet/sing-box/blob/dev-next/experimental/libbox/service.go#L148) and platform-specific behavior. I think not using something like this might break tunneling on Android?

client/service/service.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/sagernet/sing/service"
+	"github.com/sagernet/sing/service/pause"
+
+	"github.com/getlantern/radiance/client/boxoptions"
+	"github.com/getlantern/radiance/protocol"
+)
+
+// BoxService is a wrapper around libbox.BoxService
+type BoxService struct {
+	*libbox.BoxService
```
atavism, line 30: gomobile doesn't support exporting pointers to types from external packages. That means it will skip this field and any methods from libbox (like `Start`) won't be accessible in the generated bindings. You can confirm this by running..

```bash
GOOS=ios gomobile bind -v  -target ios -trimpath -ldflags="-w -s" -o bin/LanternCore.xcframework ./client/service
cat bin/LanternCore.xcframework/ios-arm64_x86_64-simulator/LanternCore.framework/Headers/Boxservice.objc.h | grep "skipped field"

// skipped field BoxService.BoxService with unsupported type: *github.com/sagernet/sing-box/experimental/libbox.BoxService
```

PR #40, opened 2025-03-18
PlatformInterface wrapper

garmr-ulfr:


client/service/service_mobile.go
```diff
@@ -0,0 +1,23 @@
+//go:build gomobile
```
garmr-ulfr, line 1: @atavism I don't know if gomobile uses a build tag so this might not be correct.

WendelHime (approved):
LGTM!

client/service/service_mobile.go
atavism: Are you able to install gomobile on your machine and test with a device?

garmr-ulfr: Yeah, I can try.

myleshorton: Note that there are a lot of gomobile build targets in sing-box...one sec

myleshorton: `make lib_ios`, for example

myleshorton: actually, this is one of the reasons I think it could make sense to use a minimal fork

garmr-ulfr: I changed the build tags to match OS. I apparently need Xcode to build for iOS. I'm going through installing/configuring everything so I can try for android.

myleshorton (approved):

PR #41, opened 2025-03-19
Use sing-box fork instead of reflect and unsafe

myleshorton:


radiance_test.go
```diff
@@ -131,7 +132,7 @@ func (msc *mockStreamConn) Write(p []byte) (n int, err error) {
 
 func TestNewRadiance(t *testing.T) {
 	t.Run("it should return a new Radiance instance", func(t *testing.T) {
-		r, err := NewRadiance(nil)
```
garmr-ulfr, line 134: Just FYI, I used `nil` here because `PlatformInterface` isn't used on desktop.

myleshorton: ok cool yeah the challenge is that the normal libbox constructor will panic [here](https://github.com/getlantern/sing-box-minimal/blob/lantern-main-next/experimental/libbox/service.go#L74)

myleshorton: I'll fiddle with that

garmr-ulfr: I already fixed that in the fork. You just have to update radiance to the new one.

client/service/service.go
```diff
@@ -57,130 +44,9 @@ func New(logOutput string, platIfce libbox.PlatformInterface) (*BoxService, erro
...
-	bs.pauseAccess.Lock()
-	defer bs.pauseAccess.Unlock()
-
-	if !bs.pauseManager.IsNetworkPaused() {
-		return
-	}
-	bs.pauseManager.NetworkWake()
-	bs.pauseTimer.Stop()
-	bs.pauseTimer = nil
-}
```
garmr-ulfr, line 102: We'll still need this. At least on desktop. It doesn't look like [libbox](https://github.com/getlantern/sing-box-minimal/blob/lantern-main-next/experimental/libbox/service_pause.go) actually pauses anything. It just calls `ResetNetwork` after 3s/3m. Maybe that's enough for mobile??

garmr-ulfr (commented):
So much cleaner than unsafe! Just one comment.

client/service/service.go
myleshorton: Thanks for reviewing! I'll add that all back. The pause call is a little weird to me -- do we actually want it to pass a duration?

I'm on it.

garmr-ulfr: > do we actually want it to pass a duration

I thought it would be useful, but we don't have to. We could just pause it until resume is called. Nord lets you pause for a duration, after which it auto-reconnects (which I actually use a lot if the site blocks VPNs so I don't forget to resume). But, they only let you pick either 1m, 5m, or 1h. I don't know about other VPNs.

garmr-ulfr: As discussed on slack, we're going to wait until maybe version 2 to support this and for now, just remove it.

garmr-ulfr (approved):

garmr-ulfr (approved):
LGTM!

myleshorton:
Thanks @garmr-ulfr!

PR #44, opened 2025-03-21
remove unnecessary test, skip incorrect tests

garmr-ulfr:


radiance.go
```diff
@@ -129,102 +117,6 @@ func (r *Radiance) SetPreferredServer(ctx context.Context, country, city string)
 	r.confHandler.SetPreferredServerLocation(country, city)
 }
 
-// Run starts the Radiance proxy server on the specified address.
-// This function will be replaced by StartVPN as part of https://github.com/getlantern/engineering/issues/1883
-func (r *Radiance) run(addr string) error {
-	reporting.Init()
```
myleshorton, line 135: I do think we still want to call this somewhere if you haven't already done that

myleshorton (approved):
Just the one comment on reporting.Init but otherwise LGTM

WendelHime (approved):

radiance.go
garmr-ulfr: I have not. `run` was never called anymore so it's a good thing you caught this!

garmr-ulfr: Added it to the constructor.

PR #47, opened 2025-03-24
Use slightly different directories depending on platform and pass them to fronted

myleshorton:
The previous setup would use a generic directory under HOME for all platforms for both data and logs, but we want things like `~/Library/Logs/Lantern` on MacOS and `%AppData%` on Windows.


myleshorton:
Pulling this in...

PR #54, opened 2025-03-30
Initial pass at setting up metrics

myleshorton:
This adds otel metrics, tracing, etc and instruments `net.Conn`s with metrics tracking.

Note that at least some of this will ultimately be in sing-box-extensions, but I'll migrate it over when the dust settles a bit!


garmr-ulfr:
Are we only tracking metrics for inbounds? Or are we just adding it to inbounds in the initial pass?

myleshorton:
> Are we only tracking metrics for inbounds? Or are we just adding it to inbounds in the initial pass?

So this is actually intended purely for server-side metrics, so only the inbounds. The next step here will be to move a lot of this over to sing-box-extensions.

PR #55, opened 2025-03-31
Embed client

garmr-ulfr:
This pull request embeds `VPNClient` in `Radiance` to improve clarity and maintainability by reducing redundant code.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -89,6 +109,33 @@ func (c *vpnClient) Stop() error {
...
 
+// ConnectionStatus returns the connection status of the VPN client
+func (c *vpnClient) ConnectionStatus() bool {
+	clientMu.Lock()
+	defer clientMu.Unlock()
+	return c.started && c.connected
+}
+
+func (c *vpnClient) setConnectionStatus(connected bool) {
+	c.connected = connected
```
WendelHime, line 120: since we're updating the `c.connnected` we probably need to lock 
```suggestion
        c.clientMu.Lock()
        defer clientMu.Unlock()
	c.connected = connected
```

WendelHime (approved):
LGTM!

client/client.go
garmr-ulfr: Thanks! I didn't actually use that function because of the strange goroutine so I completely missed that.

PR #56, opened 2025-04-03
Use common JSON config response

myleshorton:
This pull request includes significant changes to the `config` package, focusing on refactoring and improving the configuration handling and fetching process. The most important changes include renaming headers, modifying the `ConfigHandler` and related methods, and updating the `fetcher` to use JSON instead of protobuf.

### Refactoring and Improvements:

* [`backend/headers.go`](diffhunk://#diff-e07fb405996bcf124cc1c767f6851c1f8ff05caee363cdd6de36cfb9ab43c574L23-R47): Renamed `deviceIdHeader` to `deviceIDHeader` and `userIdHeader` to `userIDHeader` for consistency. Updated `NewRequestWithHeaders` and `NewIssueRequest` to use `user.BaseUser` instead of `*user.User`.
* [`config/config.go`](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R8): Removed the `Config` alias and `configResult` struct. Added `configListeners` and `configListenersMu` to `ConfigHandler`. Introduced methods `AddConfigListener` and `notifyListeners` to manage configuration listeners. Updated `fetchConfig` to handle new configuration responses and notify listeners. Changed `loadConfig` and `saveConfig` to use JSON instead of protobuf. [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R8) [[2]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L19-R21) [[3]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L36-L44) [[4]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R49-R50) [[5]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R75-R97) [[6]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R129-R179) [[7]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L185-R207) [[8]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L216-R238) [[9]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L232-R262)
* [`config/config_test.go`](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82L4-L135): Updated tests to reflect changes in `ConfigHandler` and the switch from protobuf to JSON. Added new test cases for `NewConfigHandler`.
* [`config/fetcher.go`](diffhunk://#diff-7a29ad907addd7ddb75133a0c5098cc474b34f725c8fb7baff6c2f6708fe494fR6-R16): Modified `fetcher` to use `user.BaseUser` instead of `*user.User`. Updated `fetchConfig` to use JSON for requests and responses instead of protobuf. [[1]](diffhunk://#diff-7a29ad907addd7ddb75133a0c5098cc474b34f725c8fb7baff6c2f6708fe494fR6-R16) [[2]](diffhunk://#diff-7a29ad907addd7ddb75133a0c5098cc474b34f725c8fb7baff6c2f6708fe494fL27-R49) [[3]](diffhunk://#diff-7a29ad907addd7ddb75133a0c5098cc474b34f725c8fb7baff6c2f6708fe494fL73-R64) [[4]](diffhunk://#diff-7a29ad907addd7ddb75133a0c5098cc474b34f725c8fb7baff6c2f6708fe494fL86-R76)


config/config.go
```diff
@@ -105,59 +95,83 @@ func (ch *ConfigHandler) SetPreferredServerLocation(country, city string) {
...
 
+	existingConfig, _ := ch.config.Get(eventual.DontWait)
+	if existingConfig != nil {
+		mergedConfig := existingConfig.(*C.ConfigResponse)
+		if err := mergo.MergeWithOverwrite(mergedConfig, cfg); err != nil {
+			slog.Error("merging config", "error", err)
+			return fmt.Errorf("merging config: %w", err)
+		}
+		cfg = mergedConfig
+	}
```
garmr-ulfr, line 171: Did we decide how we're going to handle un-assigning proxies? If the user has a config for a proxy that's no longer valid and they don't receive an updated version, that config will never be removed and radiance will keep testing them.

config/config_test.go
```diff
@@ -1,135 +1,211 @@
...
+			if tt.configResult != nil {
+				ch.config.Set(tt.configResult)
+			}
+
+			ctx := context.Background()
+			if tt.expectedErr == context.DeadlineExceeded {
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithTimeout(ctx, 1*time.Millisecond)
+				defer cancel()
+				time.Sleep(2 * time.Millisecond) // Ensure context expires
```
garmr-ulfr, line 53: If you pass `time.Now().Add(-1*time.Second)` to `context.WithDeadline`, it returns an expired context.

config/config.go
myleshorton: Ah great catch! I think we do want them to be able to be removed. I'll fix that.

config/config.go
```diff
@@ -54,19 +47,8 @@ type ConfigHandler struct {
...
-	City    string
-}
-
-type AvailableServerLocation struct {
-	City        string
-	Country     string
-	CountryCode string
-	Latitude    *float32
-	Longitude   *float32
+	configListeners         []func(*C.ConfigResponse)
```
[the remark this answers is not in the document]
myleshorton: Yeah this is a key issue...hmmnn...tempting to have different listeners for different parts of the config or something. I might hold off on this for now, as I feel like what to do will become clearer once we implement ads and country-specific routing rules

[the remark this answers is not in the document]
garmr-ulfr: One situation it would be useful (necessary??) is to know which outbounds/endpoints need to be updated. Not every outbound/endpoint store the config that they were created with.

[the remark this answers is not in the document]
myleshorton: Ah that's a good point - they don't know their old config. Let me do something special for outbounds and endpoints at the very least.

config/config.go
myleshorton: OK this should be fixed -- used "Override" vs "Overwrite" and added a test.

client/service/service.go
```diff
@@ -72,27 +83,17 @@ func (bs *BoxService) Start() error {
...
-	bs.pauseManager = service.FromContext[pause.Manager](ctx)
-
-	if err = lb.Start(); err == nil {
-		bs.isRunning = true
+	if err := bs.libbox.Start(); err != nil {
+		return fmt.Errorf("error starting libbox service: %w", err)
 	}
-	return err
+	bs.isRunning = true
+	return nil
```
garmr-ulfr, line 96: FYI, we need to create a new box instance in order to start it again if it was stopped. It keeps an internal state that we can't reset. [Ticket ref](https://github.com/getlantern/engineering/issues/1994)

myleshorton: Got it. I'm going to take another pass at cleaning this up.

myleshorton:
OK I think this is ready @garmr-ulfr and @WendelHime. I ended up fixing things in the general area, such as the preferred server location not being stored on disk, but let me know what you think. I also left the listener for config changes as a no-op in `service.go`, but it gets both the old config and the new one, which has also been merged with the old by the time the listener is called.

garmr-ulfr:
Reviewing now.

config/config.go
```diff
@@ -178,70 +243,83 @@ func (ch *ConfigHandler) fetchLoop(pollInterval time.Duration) {
...
+	if err := os.MkdirAll(filepath.Dir(ch.configPath), 0o755); err != nil {
+		slog.Error("creating config directory", "error", err)
+		return
 	}
+	// Marshal the config to bytes
+	// and write it to the config file.
+	// If the config is nil, we don't write anything.
+	// This is important because we don't want to overwrite the config file with an empty file.
+
+	buf, err := json.Marshal(cfg)
```
garmr-ulfr, line 289: We'll need to use the sing-box json marshaller for the sing-box config so it's marshalled correctly.

config/config.go
```diff
@@ -178,70 +243,83 @@ func (ch *ConfigHandler) fetchLoop(pollInterval time.Duration) {
...
+func (ch *ConfigHandler) GetConfig() (*Config, error) {
+	cfgRes, err := ch.config.Get(eventual.DontWait)
 	if err != nil {
-		return nil, fmt.Errorf("unmarshaling config: %w", err)
+		return nil, fmt.Errorf("getting config: %w", err)
+	}
+	cfg, ok := cfgRes.(*Config)
+	if !ok || cfg == nil {
+		return nil, fmt.Errorf("config is nil")
 	}
```
garmr-ulfr, line 309: Nit: technically, not having received the config yet is a valid path so we should return `nil, nil`. But, it's fine this way, if you prefer. Up to you!

radiance.go
```diff
@@ -170,17 +172,18 @@ func (r *Radiance) GetActiveServer() (*Server, error) {
...
 
-	return &Server{
-		Address:  config.GetAddr(),
-		Location: ServerLocation(*config.GetLocation()),
-		Protocol: config.GetProtocol(),
-	}, nil
+	log.Debug("Retrieving active server")
+	// Check if the active server is of type *Server
+	if as, ok := activeServer.(*Server); !ok {
+		return nil, fmt.Errorf("active server is not of type *Server")
```
garmr-ulfr, line 183: We should force assert here too.

config/config.go
```diff
@@ -178,70 +243,83 @@ func (ch *ConfigHandler) fetchLoop(pollInterval time.Duration) {
...
-	return &cfg, nil
+	return cfg, nil
 }
 
-// saveConfig saves the configuration to the disk.
-func saveConfig(path string, cfg *Config) error {
-	buf, err := protojson.Marshal(cfg)
+// saveWithConfig saves the config to the disk with the given config. It creates the config file
+// if it doesn't exist.
+func (ch *ConfigHandler) saveWithConfig(fn func(cfg *Config)) {
```
garmr-ulfr, line 315: What's the purpose of passing a function here?

client/service/service.go
```diff
@@ -174,3 +178,19 @@ func (bs *BoxService) Wake() {
 func (bs *BoxService) Ctx() context.Context {
 	return bs.ctx
 }
+
+// OnNewConfig is called when a new configuration is received. It updates the VPN client with the
+// new configuration and restarts the VPN client if necessary.
+func (bs *BoxService) OnNewConfig(oldConfigRaw, newConfigRaw *config.Config) error {
+	slog.Debug("Received new config")
+
+	return nil
```
garmr-ulfr, line 187: @WendelHime, Harry, and I have been returning a "not implemented" error for functions that need to be implemented so it's more likely we'll catch them in testing.

client/service/service.go
```diff
@@ -174,3 +178,19 @@ func (bs *BoxService) Wake() {
...
+}
+
+func (bs *BoxService) ParseConfig(configRaw []byte) (*config.Config, error) {
```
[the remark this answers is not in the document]
myleshorton: Yeah that's the key tradeoff between this and the earlier approach that passed the context around. I slightly prefer this only because I think the downside of having some sort of stale context is potentially much greater/unknown vs to me the lessor annoyance of having `BoxService` know about the `Config`

config/config.go
myleshorton: I think we're likely to have a lot more things that modify the config going forward, so I wanted to create a more generic approach for it. Can kill if it's too annoying =).

client/service/service.go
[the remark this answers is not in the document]
garmr-ulfr: Yeah..requiring the ctx for unmarshaling is quite annoying..

config/config.go
garmr-ulfr: It's just kind of strange/confusing that a save function needs a function passed to it. I've never seen that before. It seems out of place here. However, I do agree that we should make the config handle/config more generic, especially if radiance might be used as a VPN plugin for Segment. 

We could make `ConfigResponse`, or even `Config`, an interface and move some of config specific function out to that??

myleshorton: Yeah maybe it's misnamed? It's really just a generic modification function -- I'll rename it.

config/config.go
myleshorton: Yeah I'd just worry more in that case about the caller not checking for nil -- more confident in the caller checking for an error.

garmr-ulfr: Valid point.

garmr-ulfr (approved):
LGTM!

PR #57, opened 2025-04-09
add ability to update outbounds and endpoints during sing-box runtime

garmr-ulfr:
This pull request introduces significant updates to the `client/service/service.go` and adds new tests in `client/service/service_test.go`. The primary focus is on enhancing the management of outbounds and endpoints with new functions and tests.

### Enhancements to Outbounds and Endpoints Management:

* Introduced `updateOutboundsEndpoints` function to update outbounds and endpoints in the router, skipping any permanent ones and collecting errors encountered during the process.
* Added `updateOutbounds` and `updateEndpoints` functions to synchronize the respective managers with the provided configurations, handling exclusions, removals, and updates.
* Implemented helper functions `filterItems` and `removeItems` to facilitate the filtering and removal of items based on tags.

### Testing Enhancements:

* Added new test cases in `client/service/service_test.go` to verify the functionality of `updateOutbounds` and `updateEndpoints` functions, including scenarios for upserting, excluding, and handling errors for missing tags.

### DIFF
--- a/client/service/service.go
+++ b/client/service/service.go
@@ -14,11 +14,15 @@ import (
 	"fmt"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"sync"
 	"time"
 
 	box "github.com/sagernet/sing-box"
+	"github.com/sagernet/sing-box/adapter"
 	"github.com/sagernet/sing-box/experimental/libbox"
+	"github.com/sagernet/sing-box/log"
+	"github.com/sagernet/sing-box/option"
 	"github.com/sagernet/sing/service"
 	"github.com/sagernet/sing/service/pause"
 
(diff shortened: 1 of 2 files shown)


garmr-ulfr:
@myleshorton, @WendelHime, I'm having a hard time figuring out where this code should actually live. If either of you have a better idea, let me know and I'll move it.

myleshorton:
The location seems fine to me. I'm not 100% sold on the fanciness of the functional approach here, but I'm also unclear on how important it is to not just remove all the old ones and add the ones (how bad is it to just kill an outbound and add the same one back in the case of duplicates?).

If not that bad, something like this for outbounds and similar for inbounds certainly seems like less code:

```go
	for _, it := range outboundMgr.Outbounds() {
		if slices.Contains(excludeTags, it.Tag()) {
			continue
		}
		err := outboundMgr.Remove(it.Tag())
		if err != nil {
			errs = errors.Join(errs, fmt.Errorf("remove [%v]: %w", it.Tag(), err))
		}
	}

	// Add all the new outbounds
	for tag, outbound := range outbounds {
		err := outboundMgr.Create(ctx, router, logFactory.NewLogger(fmt.Sprintf("outbound/%s[%s]", outbound.Type, tag)), tag, outbound.Type, outbound.Options)
		if err != nil {
			errs = errors.Join(errs, fmt.Errorf("initialize [%v]: %w", tag, err))
		}
	}
```

client/service/service.go
```diff
@@ -174,3 +178,170 @@ func (bs *BoxService) Wake() {
 func (bs *BoxService) Ctx() context.Context {
 	return bs.ctx
 }
+
+var (
+	permanentOutbounds = []string{
+		"direct",
+		"dns",
+		"block",
```
myleshorton, line 186: Do we definitely not want to be able to change the config for any of these, or are they just inherently static?

garmr-ulfr:
>  I'm not 100% sold on the fanciness of the functional approach here,

Not going to lie, I hate that I had to do that to remove a lot of duplicated code..

> how bad is it to just kill an outbound and add the same one back in the case of duplicates?

Mostly performance, depending on how many we have. Also, if we remove all of them, then use user would have no connection until a new one is created and connects without issue. If, for some reason, the configs have missing or incorrect values (maybe due to bad update to the config server), none would be created. Sing-box verifies that there are no issues when creating them before it removes the old version.

At some point, we could add a grace period before removing unassigned outbounds so the transition to the new configs is more seamless and doesn't interrupt the users connection.

That being said, we can just remove all and then create the new ones for now and worry about the other stuff later.

client/service/service.go
garmr-ulfr: These 3 don't actually have config options and we'll always need them. (I think @WendelHime actually discovered that we might not need the DNS outbound, but doesn't hurt). I'm sure there will be others well add, like the `Select`/`URLTest` containers for autoselect, etc.

client/service/service.go
```diff
@@ -174,3 +178,170 @@ func (bs *BoxService) Wake() {
...
+			removeTags = append(removeTags, tag)
+		}
+	}
+	var errs error
+	for _, tag := range removeTags {
+		if err := remove(tag); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("remove [%v]: %w", tag, err))
+		}
+	}
+	return errs
```
myleshorton, line 346: Couldn't the above just be:

```go
	var errs error
	for _, it := range items {
		tag := getTag(it)
		if _, ok := newItems[tag]; !ok && !slices.Contains(excludeTags, tag) {
			if err := remove(tag); err != nil {
				errs = errors.Join(errs, fmt.Errorf("remove [%v]: %w", tag, err))
			}
		}
	}
	return errs
```

myleshorton (approved):
OK just one potential simplification but otherwise LGTM!

client/service/service.go
garmr-ulfr: I tried that, but `items` is being updated in-place and when something is removed, all the following items are shifted up. So removing `item[i]` resulted in item on the next iteration being `item[i+2]` of the original slice. I rewrote it though combining the loops.

garmr-ulfr:
This closes [#2057](https://github.com/getlantern/engineering/issues/2057)

PR #59, opened 2025-04-11
Actually update outbounds and endpoints

myleshorton:
This just updates the outbounds and endpoints when we receive a new config.


garmr-ulfr (approved):
I was just about to create a PR for this too :laughing:. LGTM!

myleshorton:
I spent some time trying to generate a better test, but I couldn't get the darn thing to fully initialize libbox

garmr-ulfr:
Better test for what?

PR #61, opened 2025-04-14
Expose multiple methods

jigar-f:
This pull request introduces several enhancements to the `Radiance` struct and related methods in the `radiance.go` file. The primary changes include adding an `httpClient` field to the `Radiance` struct, initializing it in the `NewRadiance` function, and adding new methods to expose the `httpClient` and retrieve the user object.

Key changes:

* Added `httpClient` field to the `Radiance` struct to manage HTTP client configurations.
* Initialized the `httpClient` field in the `NewRadiance` function using `k.NewHTTPClient()`.
* Added `User` method to the `Radiance` struct to return the user object associated with the client.
* Added `HttpClient` method to the `Radiance` struct to expose the HTTP client used by the Radiance client.

### DIFF
--- a/radiance.go
+++ b/radiance.go
@@ -66,6 +66,7 @@ type Radiance struct {
 	logsDir       string
 	shutdownFuncs []func(context.Context) error
 	closeOnce     sync.Once
+	httpClient    *http.Client
 }
 
 // NewRadiance creates a new Radiance VPN client. platIfce is the platform interface used to
@@ -131,6 +132,7 @@ func NewRadiance(opts client.Options) (*Radiance, error) {
 		issueReporter: issueReporter,
 		logsDir:       logDir,
 		shutdownFuncs: shutdownFuncs,
+		httpClient:    k.NewHTTPClient(),
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


radiance.go
```diff
@@ -178,6 +180,18 @@ func (r *Radiance) GetActiveServer() (*Server, error) {
...
+}
+
+// User returns the user object for this client
+func (r *Radiance) User() *user.User {
+	return r.user
+}
+
+// httpClient returns the HTTP client used by the Radiance client.
+// Expose http client
+func (r *Radiance) HttpClient() *http.Client {
```
garmr-ulfr, line 193: Why do we need to return expose an HTTP client?

garmr-ulfr: Found the discussion on [Slack](https://wdynhnkxvsdx.slack.com/archives/C04T8M4AB1Q/p1744114942489149)

jigar-f: @garmr-ulfr, update my PR.

myleshorton:
killed unused http-client

myleshorton (approved):

PR #64, opened 2025-04-15
Purchase flow

jigar-f:
This PR includes apI for purchase flow and expose that api to the client.
Client Side PR-:https://github.com/getlantern/lantern-outline/pull/44


common/headers.go
garmr-ulfr, line 1: Some of the headers and a couple of functions are already defined in `backend/headers.go`. Maybe it makes more sense for them to be here?? I'll leave that up to you, but would you consolidate these files?

common/webclient.go
```diff
@@ -1,19 +1,24 @@
...
-	"io"
+	"encoding/json"
+	"time"
+	"unicode"
 
 	"fmt"
 	"net/http"
 
+	"github.com/go-resty/resty/v2"
+	"github.com/moul/http2curl"
```
garmr-ulfr, line 13: Do we need to use these? I'm not familiar with them, but it seems like everything in this file is very easily done with `net/http`.

jigar-f: Resty provides easy access to some methods, such as hookMethod, and less boilerplate code for HTTP. But if you want, I can remove that.

jigar-f: http2curl is jsut util method to print curls is only debugging

common/headers.go
jigar-f: Thanks, I didn't know this. I use that and remove my code, Thanks for letting me know

common/webclient.go
garmr-ulfr: > Resty provides easy access to some methods, such as hookMethod, and less boilerplate code for HTTP. But if you want, I can remove that.

It's fine! I think we should only add dependencies as needed, but we probably don't need to worry about it for now.

common/user_config.go
```diff
@@ -0,0 +1,120 @@
...
+	GetUserData() (*protos.LoginResponse, error)
+	ReadSalt() ([]byte, error)
+	WriteSalt([]byte) error
+}
+
+type userConfig struct {
+	deviceID string
+	resp     *protos.LoginResponse
+	dataDir  string
+}
```
garmr-ulfr, line 30: How is this different than `user.User`? Do we need both?

common/platform.go
garmr-ulfr, line 1: Do we need this? Only `IsAndroid` and `IsIOS` is used, and they're only used once.

common/user_config.go
garmr-ulfr: Oh, I see now. `user.User` is handling communication with the API backend.

garmr-ulfr:
Looking good so far! Now that there's more than one API handler/flow, can we move the API code into it's own package (`api`)?

common/platform.go
jigar-f: I am happy to remove it, I just added this so that we might need them in the future if so, we can use that across. Let me know.

jigar-f:
> Looking good so far! Now that there's more than one API handler/flow, can we move the API code into it's own package (`api`)?

Not sure what you mean here? We have an API in each different package, one for pro and one for user.

common/platform.go
garmr-ulfr: Yeah, let's remove it for now and we can add them if needed. `IsIOS` is a bit cleaner than `runtime.GOOS == "ios"`, but just trying to avoid radiance turning into flashlight.

jigar-f: I agree, I will remove it.

jigar-f:
@garmr-ulfr What if we do something like, so we don't clutter api.go file, woulde that be fine? 
```
radiance/
├── api/
│   ├── protos/
│   ├── pro_api.go
│   ├── pro.go
│   ├── pro_test.go
│   ├── srp.go
│   ├── user_api.go
│   ├── user.go
│   └── user_test.go
├── app/
├── backend/
...
```

garmr-ulfr:
> @garmr-ulfr What if we do something like, so we don't clutter api.go file, woulde that be fine?
> 
> ```
> radiance/
> ├── api/
> │   ├── protos/
> │   ├── pro_api.go
> │   ├── pro.go
> │   ├── pro_test.go
> │   ├── srp.go
> │   ├── user_api.go
> │   ├── user.go
> │   └── user_test.go
> ├── app/
> ├── backend/
> ...
> ```

Good point, that would be great!

jigar-f:
@garmr-ulfr This is ready for final review.

api/pro_api.go
```diff
@@ -0,0 +1,92 @@
...
+	}, &resp)
+	if err != nil {
+		log.Fatalf("Error in SubscriptionPaymentRedirect: %v", err)
+		return nil, err
+	}
+	log.Printf("SubscriptionPaymentRedirect response: %v", resp)
+	return resp, nil
+}
+
+func (c *proClient) UserCreate(ctx context.Context) (*protos.UserDataResponse, error) {
```
garmr-ulfr, line 43: Can we rename this to `CreateUser` so it's more conventional?

api/pro_api.go
garmr-ulfr, line 1: With the exception of `UserCreate`, this all seems more related to just subscription handling than anything user related. I think we should change `UserCreate` to accept the user info and instead of `Pro`/`ProClient`, this can be updated to just `Subscriptions`.

radiance.go
```diff
@@ -60,7 +63,9 @@ type Radiance struct {
 	activeServer *atomic.Value
 	stopChan     chan struct{}
 
-	user *user.User
+	user       *api.User
+	pro        *api.Pro
```
garmr-ulfr, line 67: Can we combined all API communication/handling into a single `APIClient` handler?

garmr-ulfr (changes requested):
Can you add documentation to functions and structs (that you added) following godoc format: 
```
<func/struct name> does a thing 
```
it only needs to be enough that someone using that function would know exactly how to use it and what it's for without having to read the source code.

api/pro_api.go
```diff
@@ -0,0 +1,92 @@
...
+	SubscriptionPaymentRedirect(ctx context.Context, data *protos.SubscriptionPaymentRedirectRequest) (*protos.SubscriptionPaymentRedirectResponse, error)
+	StripeSubscription(ctx context.Context, data *protos.SubscriptionRequest) (*protos.SubscriptionResponse, error)
+	UserCreate(ctx context.Context) (*protos.UserDataResponse, error)
+	Plans(ctx context.Context) (*protos.PlansResponse, error)
+}
+
+func (c *proClient) SubscriptionPaymentRedirect(ctx context.Context, data *protos.SubscriptionPaymentRedirectRequest) (*protos.SubscriptionPaymentRedirectResponse, error) {
+	var resp *protos.SubscriptionPaymentRedirectResponse
+
+	err := c.Get(ctx, "/subscription-payment-redirect", map[string]any{
```
WendelHime, line 28: Since we're using the proto for defining the response, I guess we could use the request struct instead of a map

api/protos/response.go
WendelHime, line 1: We could probably rename this file to subscription too

WendelHime (commented):
Added a few suggestions! Looks great!

api/pro_api.go
WendelHime: Wait we're already receiving a proto request

api/protos/auth.proto
```diff
@@ -143,3 +145,38 @@ message LogoutRequest {
...
+//   repeated Device device = 8;
+//   string code = 9;
+//   int64 expireAt = 10;
+//   string referral = 11;
+//   string token = 12;
+//   bool yinbiEnabled = 13;
+//   repeated string inviters = 14;
+//   repeated string invitees = 15;
+//   repeated Purchase purchases = 16;
+// }
```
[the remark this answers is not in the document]
jigar-f: removed

common/user_config.go
```diff
@@ -0,0 +1,120 @@
...
+	data, err := os.ReadFile(readPath)
+	if err != nil {
+		return nil, err
+	}
+	var resp protos.LoginResponse
+	if err := proto.Unmarshal(data, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
+}
```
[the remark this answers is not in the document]
jigar-f: Done

common/webclient.go
```diff
@@ -22,16 +28,50 @@ type WebClient interface {
...
+	Post(ctx context.Context, path string, params map[string]any, target any) error
 }
 
 type webClient struct {
-	*http.Client
+	*resty.Client
+}
+
+// Opts are common options that RESTClient may be configured with
+type Opts struct {
```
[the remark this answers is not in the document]
jigar-f: Change to WebClientOptions

common/webclient.go
```diff
@@ -40,37 +80,24 @@ func NewWebClient(httpClient *http.Client) WebClient {
...
+		req.SetQueryParams(convertToStringMap(params))
 	}
-	req.Header.Set("Content-Type", "application/x-protobuf")
-	req.Header.Set("Accept", "application/x-protobuf")
-
-	resp, err := c.Do(req)
+	//Overide the default content type
+	// to application/x-protobuf
+	req.Header.Set("Content-Type", ContentTypeJSON)
+	req.Header.Set("Accept", ContentTypeJSON)
```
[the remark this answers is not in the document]
jigar-f: Thanks for catching, fixed.

common/webclient.go
```diff
@@ -83,28 +110,102 @@ func (c *webClient) PostPROTOC(ctx context.Context, path string, msg, target pro
...
+	if params != nil {
+		req.SetQueryParams(convertToStringMap(params))
+	}
+	req.Header.Set("Content-Type", ContentTypeJSON)
+	req.Header.Set("Accept", ContentTypeJSON)
+
+	resp, err := req.Get(path)
+
+	command, _ := http2curl.GetCurlCommand(req.RawRequest)
+	fmt.Printf("curl command: %v", command)
```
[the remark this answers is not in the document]
jigar-f: Sure, I'll remove it. What do you think about having a dev variable? If it's dev, we print the curl value, and if not, we do not. That really helps, and we don't need to add/remove curl again when debugging.

common/webclient.go
```diff
@@ -83,28 +110,102 @@ func (c *webClient) PostPROTOC(ctx context.Context, path string, msg, target pro
...
+	if params != nil {
+		req.SetBody(params)
 	}
+	req.Header.Set("Content-Type", ContentTypeJSON)
+	req.Header.Set("Accept", ContentTypeJSON)
+
+	resp, err := req.Post(path)
+
+	command, _ := http2curl.GetCurlCommand(req.RawRequest)
+	fmt.Printf("curl command: %v", command)
```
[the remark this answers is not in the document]
jigar-f: Done

api/pro.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/getlantern/radiance/app"
+	"github.com/getlantern/radiance/backend"
+	"github.com/getlantern/radiance/common"
+	"github.com/go-resty/resty/v2"
+)
+
+// Pro represents pro server apis
+type Pro struct {
+	proClient ProClient
+}
```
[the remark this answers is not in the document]
jigar-f: Ah, for now, I think we only added the subscription part — that’s why we just have the subscription API. But as we add more features, we’ll definitely have more APIs here, like `GetUserData`, `RestorePurchase`, etc.

That said, if you still think we should move to `ProClient`, let me know.

Just FYI:  
- All `/pro-server` APIs go into `pro_api.go`  
- All `/user-auth` APIs go into `user_api.go`

api/pro_api.go
```diff
@@ -0,0 +1,92 @@
...
+		"provider":         data.Provider,
+		"plan":             data.Plan,
+		"deviceName":       data.DeviceName,
+		"email":            data.Email,
+		"subscriptionType": data.SubscriptionType,
+	}, &resp)
+	if err != nil {
+		log.Fatalf("Error in SubscriptionPaymentRedirect: %v", err)
```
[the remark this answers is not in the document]
jigar-f: Thanks @WendelHime, changed all `log` calls with `slog`.

api/pro_test.go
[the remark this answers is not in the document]
jigar-f: I have updated the tests to use the new changes.

api/pro_api.go
jigar-f: Done

api/pro_api.go
jigar-f: Same as above.

common/user_config.go
```diff
@@ -0,0 +1,120 @@
...
+	WriteSalt([]byte) error
+	Locale() string
+}
+
+type userConfig struct {
+	deviceID string
+	resp     *protos.LoginResponse
+	dataDir  string
+	locale   string
+}
```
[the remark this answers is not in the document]
jigar-f: Done, Changed to `UserInfo/userInfo`

common/user_config.go
```diff
@@ -0,0 +1,120 @@
...
+	resp, _ := ReadUserData(dataDir)
+	u := &userConfig{deviceID: deviceID, resp: resp, dataDir: dataDir, locale: locale}
+	return u
+}
+
+func (u *userConfig) Locale() string {
+	if u.locale != "" {
+		return u.locale
+	}
+	return ""
```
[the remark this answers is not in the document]
jigar-f: Done

radiance.go
jigar-f: you mean somethinig like this
```
type APIClient struct {
	user *api.User
	pro  *api.Pro
}
```


```
type Radiance struct {
	client.VPNClient

	confHandler  configHandler
	activeServer *atomic.Value
	stopChan     chan struct{}

	// apiClient is the API client used to interact user and pro APIs
	apiClient *APIClient
	// user config is the user config object that contains the device ID and other user data
	userConfig common.UserInfo

	issueReporter *issue.IssueReporter
	logsDir       string
	shutdownFuncs []func(context.Context) error
	closeOnce     sync.Once
}

```

api/pro_api.go
jigar-f: `Get(ctx context.Context, path string, params map[string]any, target any) error`

Get request needs map as params.

api/protos/response.go
jigar-f: @WendelHime I added a [comment](https://github.com/getlantern/radiance/pull/64#discussion_r2059561539) at top. `response.go` will have many more structs soon as we add other APIs.

jigar-f:
@WendelHime @garmr-ulfr I pushed new changes and replied to your comments. Thanks

api/pro_api.go
garmr-ulfr: Hmm.. I still feel like we should separate out the subscription api from pro and user apis. It would be much cleaner to have subscription api methods accept user info as a parameter when it's needed. This would also make it much easier to update if we ever go with a different subscription service.

api/pro.go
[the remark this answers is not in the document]
garmr-ulfr: Is `GetUserData`, `DataCapInfo`, and `PrivateServer` specific to pro users? If possible, we should de-couple things that aren't closely related, _i.e._ a pro user is a user with a subscription but not necessarily the subscription itself. So we can de-couple the subscription and pro user API and we can just pass the user info to the subscription API methods. This will keep everything much cleaner and easier to maintain.

radiance.go
garmr-ulfr: Yes! But define `APIClient` in the api package.

common/webclient.go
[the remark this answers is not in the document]
garmr-ulfr: I'm unclear on how is the curl cmd is useful, I'm sure it's something simple, but could you elaborate? 
:thinking: Instead of a dev variable, we could just log it if log level is "trace" (`log.Trace("curl: ", command)`). We only use "trace" when debugging/testing anyway. Would that work?

api/pro_api.go
jigar-f: Sure, I can make changes separate out the subscription API.

radiance.go
jigar-f: Do you mind looking at this #73 . This might what you are looking for.

api/pro_test.go
```diff
@@ -0,0 +1,60 @@
...
+		Provider:         "stripe",
+		Plan:             "pro",
+		DeviceName:       "test-device",
+		Email:            "",
+		SubscriptionType: protos.SubscriptionType("monthly"),
+	}
+	resp, error := proServer.SubscriptionPaymentRedirect(context.Background(), body)
+	assert.NoError(t, error)
+	assert.NotNil(t, resp.Redirect)
+	fmt.Printf("SubscriptionPaymentRedirect response: %v", resp)
```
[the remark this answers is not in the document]
garmr-ulfr: Same for the rest of the tests

init_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInitialize(t *testing.T) {
+	opts := client.Options{
+		DataDir:  "/tmp/data",
+		LogDir:   "/tmp/logs",
+		Locale:   "en-US",
+		DeviceID: "test-device-id",
+	}
```
garmr-ulfr, line 16: This will ensure that the files/folders are cleaned up when the test finishes.
```suggestion
func TestInitialize(t *testing.T) {
	tmp := t.TempDir()
	opts := client.Options{
		DataDir:  tmp,
		LogDir:   tmp,
		Locale:   "en-US",
		DeviceID: "test-device-id",
	}
```

garmr-ulfr:
I'll fix the merge conflicts and address the comments as they're minor.

jigar-f:
@garmr-ulfr Should I go ahed and merge this?

garmr-ulfr:
@jigar-f, yes! Sorry I meant to do that but got distracted fixing a bug!

jigar-f:
Thanks @garmr-ulfr, I'll take it for a spin last time, test it, and then merge

PR #67, opened 2025-04-17
Generating wg keys and using them

WendelHime:
Closes getlantern/engineering#2047

This pull request:
- Load wireguard key from stored file locally, if doesn't exist returns an error
- Generate a wireguard key if the file doesn't exist locally
- Use the public key when requesting configs
- Set the private key on wireguard endpoints


myleshorton (approved):
Assuming the server is always able to dynamically associate our new key with the proxy it returns, this LGTM!

myleshorton (approved):

PR #70, opened 2025-04-23
Fix directory handling when radiance is created

myleshorton:
This honors the log directory passed into options on radiance creation and uses the new config endpoint.


myleshorton:
This is trivial one -- YOLOing...

PR #72, opened 2025-04-25
Support Lantern Server Manager in Radiance

reflog:
closes https://github.com/getlantern/engineering/issues/2117

blocked at the moment by https://github.com/getlantern/radiance/pull/35

### DIFF
--- a/client/service/custom_server_manager.go
+++ b/client/service/custom_server_manager.go
@@ -31,7 +31,7 @@ func NewCustomServerManager(ctx context.Context, dataDir string) *CustomServerMa
 		customServers:             make(map[string]CustomServerInfo),
 		customServersMutex:        new(sync.RWMutex),
 		customServersFilePath:     filepath.Join(dataDir, "custom_servers.json"),
-		trustedServerFingerprints: filepath.Join(dataDir, "data", "trusted_server_fingerprints.json"),
+		trustedServerFingerprints: filepath.Join(dataDir, "trusted_server_fingerprints.json"),
 	}
 	csm.SetContext(ctx)
 	return csm
--- a/client/service/custom_server_manager.go
+++ b/client/service/custom_server_manager.go
@@ -2,8 +2,11 @@ package boxservice
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"sync"
@@ -67,7 +70,7 @@ func (m *CustomServerManager) AddCustomServer(cfg ServerConnectConfig) error {
 		return fmt.Errorf("failed to unmarshal config: %w", err)
 	}
 
-	if (loadedOptions.Endpoint == nil && loadedOptions.Outbound == nil) || loadedOptions.Tag == "" {
+	if loadedOptions.Endpoint == nil && loadedOptions.Outbound == nil {
 		return fmt.Errorf("invalid custom server provided")
 	}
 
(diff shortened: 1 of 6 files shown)


reflog:
@WendelHime @garmr-ulfr ready for review

client/service/custom_server_manager.go
```diff
@@ -282,3 +285,92 @@ func (m *CustomServerManager) newSelectorOutbound(outboundManager adapter.Outbou
 
 	return nil
 }
+
+func insecureHttpClient() http.Client {
+	transport := http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
```
myleshorton, line 292: This seems tricky in terms of the potential to for adversaries to MITM traffic...

Can we just issue trusted certs?

reflog: @myleshorton hm. and how would we do that? for that we need to be the signing authority, right? so we can't just generate self-signed certificates on lantern-server-manager startup.
Could you elaborate what did you have in mind?

myleshorton: Can't we use let's encrypt and just trust their root CAs?

reflog: @myleshorton how can we use Let's Encrypt if the user doesn't have a domain name? (which I think most people won't)

myleshorton: Ah of course got it. I wonder if we can make the public key part of the config somehow? When we create these servers, do any of them have generic domains we can use? Like doesn't AWS assign some super generic subdomain to each ec2 instance, for example?

myleshorton: > I think the only way to avoid having un-trusted TLS is - on start of Lantern Server Manager we contact Lantern API that will sign the certificate.

Makes sense. What if we did it in code on the server manager side? As in:

1. Have the server manager generate a new cert on the first run
2. Call our API to get it signed by our own trusted root CA or whatever
3. Have the client trust that root CA

myleshorton: Ha just realized that's what you're saying....yes I think that sounds right!

reflog:
@myleshorton @Crosse after our discussion, I went with the TOFU approach.
API is adjusted accordingly

client/service/tofu.go
```diff
@@ -0,0 +1,184 @@
...
+
+// getServerFingerprints connects to an untrusted server, and get cert information about it
+// the addr must be a valid address, including port
+func getServerFingerprints(addr string) ([]CertDetail, error) {
+	var ret []CertDetail
+
+	// If the port isn't specified, assume https standard port
+	if !strings.Contains(addr, ":") {
+		addr = addr + ":443"
+	}
```
Crosse, line 99: This will trigger a false positive for bare IPv6 addresses. Probably better to use [`net.SplitHostPort()`](https://pkg.go.dev/net#SplitHostPort) and check for an empty `port` return value.

client/service/tofu.go
```diff
@@ -0,0 +1,184 @@
...
+func readTrustedServerFingerprints(fingerprintsJSON string) (map[string]string, error) {
+	fingerprints := map[string]string{}
+	if _, err := os.Stat(fingerprintsJSON); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fingerprints, nil
+		}
+		return nil, fmt.Errorf("failed to check trusted server fingerprints: %w", err)
+	}
+
+	data, err := os.ReadFile(fingerprintsJSON)
```
Crosse, line 141: You can collapse the `os.Stat()` call into the `os.ReadFile()` call and avoid a potential [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) scenario.

client/service/tofu.go
```diff
@@ -0,0 +1,184 @@
...
+func writeTrustedServerFingerprints(fingerprintsJSON string, fingerprints map[string]string) error {
+	data, err := json.Marshal(fingerprints)
+	if err != nil {
+		return fmt.Errorf("failed to marshal trusted server fingerprints: %w", err)
+	}
+	if err := os.MkdirAll(filepath.Dir(fingerprintsJSON), 0755); err != nil {
+		return err
+	}
+
+	if err := os.WriteFile(fingerprintsJSON, data, 0644); err != nil {
```
Crosse, line 162: SSH uses `0600` for permissions on `~/.ssh/authorized_keys`, which may be a good idea to use here as well.

Crosse (commented):
Just a couple of comments!

reflog:
@Crosse thanks, comments addressed

reflog:
@garmr-ulfr @WendelHime @myleshorton unless anyone objects - i'll merge this

garmr-ulfr (approved):
LGTM. Just a couple of comments.

PR #81, opened 2025-05-01
fix unmarshal bug. unmarshal as configresponse instead of config.Config

garmr-ulfr:
Copilot couldn't actually describe what this PR does.

This PR changes the config parser/unmarshaller to parse as a `ConfigResponse` object, instead of a `Config` object. This fixes a bug where the box options were being overwritten with an empty one on every fetch.

### Configuration Handling Refactor:
* Added new `Options` struct in `ConfigHandler`, encapsulating all configuration options passed to `NewConfigHandler`. (`config/config.go`, [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L50-R61) [[2]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L74-R103)
* Added a `mergeResp` function to handle deep merging of old and new configuration responses, ensuring proper updates while preserving specific fields. (`config/config.go`, [config/config.goR290-R306](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R290-R306))

### Test Case Updates:
* Updated test cases to use the new `confRespParser` field and `unmarshalConfig` method in `ConfigHandler`. (`config/config_test.go`, [[1]](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R16-R22) [[2]](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82L34-R35) [[3]](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82L73-R74)
* Adjusted mock implementations and test data to align with the new configuration response structure. (`config/config_test.go`, [[1]](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82L157-R158) [[2]](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82L167-L172)

### Client Integration:
* Refactored `Radiance` to use the new `Options` struct when initializing `ConfigHandler`, improving consistency and reducing boilerplate. (`radiance.go`, [radiance.goL133-R141](diffhunk://#diff-60179a9bacb28016fc7cda79abcad6f09a5c467689999b56f2c6e0cb3d94140cL133-R141))
* Added a synchronization mechanism to the `Close` method in `Radiance` to ensure proper shutdown. (`radiance.go`, [radiance.goR179](diffhunk://#diff-60179a9bacb28016fc7cda79abcad6f09a5c467689999b56f2c6e0cb3d94140cR179))

These changes collectively enhance the robustness and maintainability of the configuration management system while simplifying integration points across the codebase.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,9 +15,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	"dario.cat/mergo"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 
 	C "github.com/getlantern/common"
+	"github.com/qdm12/reprint"
 	"github.com/sagernet/sing-box/constant"
 	"github.com/sagernet/sing-box/option"
 	singjson "github.com/sagernet/sing/common/json"
@@ -267,13 +269,13 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
 	}
 	oldConfig, _ := ch.GetConfig()
 	if oldConfig != nil {
-		merged, err := mergeResp(oldConfig.ConfigResponse, cfg.ConfigResponse)
+		merged, err := mergeResp(&oldConfig.ConfigResponse, &cfg.ConfigResponse)
 		if err != nil {
 			slog.Error("failed to merge config", "error", err)
 			return
 		}
-
 		cfg.ConfigResponse = *merged
+
 		if cfg.PreferredLocation == (C.ServerLocation{}) {
 			cfg.PreferredLocation = ch.preferredLocation.Load().(C.ServerLocation)
 		}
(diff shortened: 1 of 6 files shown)


WendelHime (approved):

myleshorton:
Still working through this..

config/config.go
```diff
@@ -254,25 +267,17 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
...
-	}
-	// If the merged config has no server location, use the preferred location
-	// stored in memory, if any.
-	if cfg.PreferredLocation == (C.ServerLocation{}) {
-		storedLocation := ch.preferredLocation.Load()
-		if storedLocation != nil {
-			cfg.PreferredLocation = storedLocation.(C.ServerLocation)
+		cfg.ConfigResponse = merged
+
+		if cfg.PreferredLocation != (C.ServerLocation{}) {
```
garmr-ulfr, line 279: Just noticed this should be `==` not `!=`

config/config.go
```diff
@@ -282,6 +287,23 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
...
+	oldConfig.Options = option.Options{}
+	newConfig.Options = option.Options{}
+
+	// badjson.Merge ensures a deep copy of the old config is made
+	merged, err := badjson.Merge(context.Background(), oldConfig, newConfig, false)
+	if err != nil {
+		return C.ConfigResponse{}, err
+	}
+	merged.Options = newOpts
+	return merged, nil
```
myleshorton, line 304: I don't think this is quite what we want. For example, it will usually be the case that we don't want to change the sing-box DNS options from the server, but with this we have to include them in the response each time.

On the other hand, occasionally the available server locations will change, and I think we should just replace those wholesale vs worrying about the exact diff between the locations the user knows about vs the new locations. 

This doesn't work in either case. 

Gonna make a PR to this PR with a test for these various cases!

config/config.go
garmr-ulfr: Fixed

config/config.go
```diff
@@ -276,7 +276,7 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
 		}
 		cfg.ConfigResponse = merged
 
-		if cfg.PreferredLocation != (C.ServerLocation{}) {
+		if cfg.PreferredLocation == (C.ServerLocation{}) {
```
myleshorton, line 279: Nice was going to mention that one =)

config/config.go
garmr-ulfr: Are we going to be updating the DNS? I thought it was just going to be the outbounds and endpoints. 

This should work for the second case, for outbounds and endpoints. If any server configs change, we have to include all of them in the response. There's no way for the client to know if, and which, configs should just be updated or which to remove, unless we include a field in the response that indicates that.

garmr-ulfr: Currently, the config API returns a `option.Options` the only includes the `Outbounds` and `Endpoint` fields. All other fields are left empty as they should never change. Changing the DNS options does make sense though. So we could include the `DNS` field as well. Except, it would only need to be included if the DNS config needs to be changed.

myleshorton: >  If any server configs change, we have to include all of them in the response. 

Right the prior code using `badjson` would combine the server locations from the old and the new, which I don't think is what we want. Mergo will overwrite new slices entirely vs combining them, which I think is what we want for all the slice fields just for sanity's sake.

garmr-ulfr:
Ignore my last 2 commit!

config/config.go
garmr-ulfr: > Right the prior code using badjson would combine the server locations from the old and the new, which I don't think is what we want.

Yeah, I realized that was a problem when going through your tests. 

> overwrite new slices entirely vs combining them, which I think is what we want for all the slice fields just for sanity's sake.

Agreed. That's exactly what we want.

garmr-ulfr: Well, between here and your PR, you've convinced me mergo is the best solution.

config/config.go
```diff
@@ -282,6 +285,49 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
...
+		}
+
+		oldConfig.Servers = newConfig.Servers
+		oldConfig.OutboundLocations = newConfig.OutboundLocations
+	}
+
+	oldConfig.Options = opts
+	return &oldConfig, nil
+}
+
```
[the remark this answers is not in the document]
garmr-ulfr: Also valid. I don't see a use case for updating routing options, but if we're already doing NTP, DNS, Outbounds, etc, we might as well do everything.

PR #84, opened 2025-05-02
Updated config merge to preserve elements we want to from old config while using new elements we want

myleshorton:
This switches back to using mergo for merging config structs so that we preserve the elements of the old config we want while using elements of the new config we want. This allows the server side to carefully include only the changes we desire vs responding the the entire sing-box options every time.

This pull request refactors the configuration merging logic in `config/config.go` to improve maintainability and correctness. It replaces the custom `mergeResp` function with the `mergo` library for merging configurations and updates the associated test suite to ensure comprehensive coverage of the new behavior.

### Refactoring and dependency updates:

* Replaced the custom `mergeResp` function with an implementation using the `mergo` library to simplify and improve the merging logic for configuration responses. The new implementation uses pointers for `ConfigResponse` and supports overriding fields during the merge. (`config/config.go`, [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L272-R278) [[2]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L292-R296)
* Added the `dario.cat/mergo` dependency to the project in `go.mod` to support the new configuration merging functionality.

### Bug fixes in configuration handling:

* Fixed a bug in `setConfigAndNotify` where the `PreferredLocation` field was incorrectly updated. The condition was inverted to ensure that the default value is only set when `PreferredLocation` is empty. (`config/config.go`, [config/config.goL272-R278](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L272-R278))

### Test suite improvements:

* Added a new test function, `TestMergeResp`, to verify the behavior of the updated `mergeResp` function. The tests cover various scenarios, including successful merges, handling of empty server locations, overwriting outbound options, and preserving DNS options. (`config/config_test.go`, [config/config_test.goR234-R411](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R234-R411))

### Code cleanup:

* Removed the unused `context` import and the deprecated `badjson` library from `config/config.go`. (`config/config.go`, [[1]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L7) [[2]](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R18-L25)
* Added a missing alias for the `option` package (`O`) in `config/config_test.go` for consistency. (`config/config_test.go`, [config/config_test.goR12](diffhunk://#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R12))

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,6 +19,7 @@ import (
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 
 	C "github.com/getlantern/common"
+	"github.com/qdm12/reprint"
 	"github.com/sagernet/sing-box/constant"
 	"github.com/sagernet/sing-box/option"
 	singjson "github.com/sagernet/sing/common/json"
@@ -289,11 +290,12 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
 // mergeResp merges the old and new configuration responses. The merged response is returned
 // along with any error that occurred during the merge.
 func mergeResp(oldConfig, newConfig *C.ConfigResponse) (*C.ConfigResponse, error) {
-	if err := mergo.Merge(oldConfig, newConfig, mergo.WithOverride); err != nil {
+	oldConfigCopy := reprint.This(*oldConfig).(C.ConfigResponse)
+	if err := mergo.Merge(&oldConfigCopy, newConfig, mergo.WithOverride); err != nil {
 		slog.Error("merging config", "error", err)
 		return newConfig, err
 	}
-	return oldConfig, nil
+	return &oldConfigCopy, nil
 }
 
 // fetchLoop fetches the configuration every pollInterval.
--- a/config/config.go
+++ b/config/config.go
@@ -4,7 +4,6 @@ Package config provides a handler for fetching and storing proxy configurations.
 package config
 
 import (
-	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -16,13 +15,13 @@ import (
 	"sync/atomic"
 	"time"
 
+	"dario.cat/mergo"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 
 	C "github.com/getlantern/common"
 	"github.com/sagernet/sing-box/constant"
 	"github.com/sagernet/sing-box/option"
 	singjson "github.com/sagernet/sing/common/json"
-	"github.com/sagernet/sing/common/json/badjson"
 
 	"github.com/getlantern/radiance/common"
 	"github.com/getlantern/radiance/user"
@@ -269,14 +268,14 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
 	}
 	oldConfig, _ := ch.GetConfig()
 	if oldConfig != nil {
-		merged, err := mergeResp(oldConfig.ConfigResponse, cfg.ConfigResponse)
+		merged, err := mergeResp(&oldConfig.ConfigResponse, &cfg.ConfigResponse)
 		if err != nil {
 			slog.Error("merging config", "error", err)
 			return
 		}
-		cfg.ConfigResponse = merged
+		cfg.ConfigResponse = *merged
 
-		if cfg.PreferredLocation != (C.ServerLocation{}) {
+		if cfg.PreferredLocation == (C.ServerLocation{}) {
 			cfg.PreferredLocation = ch.preferredLocation.Load().(C.ServerLocation)
 		}
 	}
@@ -289,19 +288,12 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
 
 // mergeResp merges the old and new configuration responses. The merged response is returned
 // along with any error that occurred during the merge.
-func mergeResp(oldConfig, newConfig C.ConfigResponse) (C.ConfigResponse, error) {
-	newOpts := newConfig.Options
-	// we set the options to empty since we want to use the new options. we replace them later
-	oldConfig.Options = option.Options{}
-	newConfig.Options = option.Options{}
-
-	// badjson.Merge ensures a deep copy of the old config is made
-	merged, err := badjson.Merge(context.Background(), oldConfig, newConfig, false)
-	if err != nil {
-		return C.ConfigResponse{}, err
+func mergeResp(oldConfig, newConfig *C.ConfigResponse) (*C.ConfigResponse, error) {
+	if err := mergo.Merge(oldConfig, newConfig, mergo.WithOverride); err != nil {
+		slog.Error("merging config", "error", err)
+		return newConfig, err
 	}
-	merged.Options = newOpts
-	return merged, nil
+	return oldConfig, nil
 }
 
 // fetchLoop fetches the configuration every pollInterval.
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 	"testing"
 
+	O "github.com/sagernet/sing-box/option"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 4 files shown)


config/config.go
```diff
@@ -289,19 +288,12 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
...
-	newConfig.Options = option.Options{}
-
-	// badjson.Merge ensures a deep copy of the old config is made
-	merged, err := badjson.Merge(context.Background(), oldConfig, newConfig, false)
-	if err != nil {
-		return C.ConfigResponse{}, err
+func mergeResp(oldConfig, newConfig *C.ConfigResponse) (*C.ConfigResponse, error) {
+	if err := mergo.Merge(oldConfig, newConfig, mergo.WithOverride); err != nil {
+		slog.Error("merging config", "error", err)
+		return newConfig, err
```
garmr-ulfr, line 294: This will overwrite the old config and then we won't be able to send it to the listeners. That was why I switched to using `sing-box/badjson`. It will merge objects into a new one without modifying the original.

config/config_test.go
```diff
@@ -230,6 +231,172 @@ func TestHandlerFetchConfig(t *testing.T) {
...
+		}
+
+		mergedConfig, err := mergeResp(oldConfig, newConfig)
+		require.NoError(t, err, "Should not return an error when merging configs")
+		assert.Equal(t, newConfig.Servers, mergedConfig.Servers, "Merged servers should match newConfig servers")
+		assert.Equal(t, newConfig.Options, mergedConfig.Options, "Merged options should match newConfig options")
+		assert.Equal(t, newConfig.UserInfo.IP, mergedConfig.UserInfo.IP, "Merged IP should match newConfig IP")
+		assert.Equal(t, oldConfig.UserInfo.ProToken, mergedConfig.UserInfo.ProToken, "Merged ProToken should match oldConfig ProToken")
+	})
+
```
garmr-ulfr, line 272: If you add this test, it fails.
```suggestion

	t.Run("SuccessfulRemovedUnassignedOutbounds", func(t *testing.T) {
		oldConfig := &C.ConfigResponse{
			UserInfo: C.UserInfo{
				IP:       "1.1.1.1",
				ProToken: "test-pro-token",
			},
			Servers: []C.ServerLocation{
				{Country: "US", City: "New York"},
			},
			Options: O.Options{
				Outbounds: []O.Outbound{
					{
						Tag: "outbound1",
						Options: map[string]interface{}{
							"key": "value",
						},
					},
					{
						Tag: "outbound4",
						Options: map[string]interface{}{
							"key": "value",
						},
					},
				},
			},
		}
		newConfig := &C.ConfigResponse{
			UserInfo: C.UserInfo{
				IP: "2.2.2.2",
			},
			Servers: []C.ServerLocation{
				{Country: "UK", City: "London"},
			},
			Options: O.Options{
				Outbounds: []O.Outbound{
					{
						Tag: "outbound2",
						Options: map[string]interface{}{
							"key": "value",
						},
					},
				},
			},
		}
		want := &C.ConfigResponse{
			UserInfo: C.UserInfo{
				IP:       "1.1.1.1",
				ProToken: "test-pro-token",
			},
			Servers: []C.ServerLocation{
				{Country: "UK", City: "London"},
			},
			Options: O.Options{
				Outbounds: []O.Outbound{
					{
						Tag: "outbound2",
						Options: map[string]interface{}{
							"key": "value",
						},
					},
				},
			},
		}

		mergedConfig, err := mergeResp(oldConfig, newConfig)
		require.NoError(t, err, "Should not return an error when merging configs")
		assert.Equal(t, mergedConfig, want)
	})
	
```

config/config.go
garmr-ulfr: Does it?  The other tests passed, but maybe they need to be expanded. Since there isn't much that needs to be updated, we could just do it manually instead of introducing more dependencies.

myleshorton: Yeah I added a bunch more tests to verify those various cases.

config/config_test.go
myleshorton: checking

myleshorton: That only fails because the IP we want should be 2.2.2.2 -- the IP from the new config

myleshorton: Just added that test in

garmr-ulfr: > That only fails because the IP we want should be 2.2.2.2 -- the IP from the new config

Oops, copied the wrong line :smile:

config/config.go
myleshorton: > Since there isn't much that needs to be updated, we could just do it manually instead of introducing more dependencies.

You mean do the merge manually? That seems potentially buggy to me, especially with backwards compatibility in mind as we inevitably introduce changes.

garmr-ulfr: > That seems potentially buggy to me, especially with backwards compatibility in mind as we inevitably introduce changes.

Hmmmm..that's a good point. Fair enough.

config/config.go
```diff
@@ -289,19 +289,13 @@ func (ch *ConfigHandler) setConfigAndNotify(cfg *Config) {
...
-	// we set the options to empty since we want to use the new options. we replace them later
-	oldConfig.Options = option.Options{}
-	newConfig.Options = option.Options{}
-
-	// badjson.Merge ensures a deep copy of the old config is made
-	merged, err := badjson.Merge(context.Background(), oldConfig, newConfig, false)
-	if err != nil {
-		return C.ConfigResponse{}, err
+func mergeResp(oldConfig, newConfig *C.ConfigResponse) (*C.ConfigResponse, error) {
+	oldConfigCopy := reprint.This(*oldConfig).(C.ConfigResponse)
```
garmr-ulfr, line 293: We could just do this to avoid adding reprint. This does a deep copy
```suggestion
	var oldCopy C.ConfigResponse
	buf, _ := json.Marshal(oldConfig)
	_ = json.Unmarshal(buf, &oldCopy)

```

myleshorton: yeah I do generally like the json way of doing it, but reprint's way underneath is slicker

config/config_test.go
myleshorton: I just added a more complicated merge case for `Options` that shows that

config/config.go
garmr-ulfr: I do feel we should try to keep dependencies to a minimum, only introducing them when necessary. But if you prefer we use reprint, then alright.

config/config_test.go
garmr-ulfr: We do need to make sure if either Outbounds or Endpoints exist that we update both to the new config. While sing-box treats them as different, for us they're the same thing, just server proxy configs.

config/config.go
myleshorton: Yeah I generally agree, but this deep copy problem is pretty common, the code for doing it is super minimal if you click through, and I think it's pretty clearly the "right" way to do it I think.

config/config_test.go
myleshorton: Oh interesting OK. Can we safely handle that on the server side or no?

config/config.go
myleshorton: I actually just emailed the developer because he also wrote his own VPN in Go and I think Flutter

config/config_test.go
garmr-ulfr: We'll need to handle that on the client. If there are new outbound configs but none for endpoints, we have to treat that as the client's been unassigned endpoint servers. And vice versa (although extremely rare).

config/config.go
garmr-ulfr: > but this deep copy problem is pretty common

I could have sworn go had a built in deep copy, I think `reflect.DeepCopy`. I spent 20 minutes yesterday searching online for where they moved it..you would think they would include that.

garmr-ulfr (approved):
LGTM

garmr-ulfr:
Merging

PR #92, opened 2025-05-09
Separate vpn

garmr-ulfr:
This pull request includes significant changes to refactor the codebase, simplify configuration handling, and improve modularity. Key updates include replacing inline outbound and endpoint definitions with reusable base structures, refactoring configuration management to use a file-based approach with a watcher, and decoupling `VPNClient` from `Radiance` so it can be run independently. 

### Configuration Management Refactor:
* Introduced `ConfigFileName` constant in `app/app.go` to standardize the configuration file name (`config.json`).
* Refactored `BoxService` to load configuration options from a file and added a file watcher to reload options dynamically when the file changes. [[1]](diffhunk://#diff-3e93cc11b22eab442b7c9435196e3101e2ac6f286ee1b63c98f0de4099bc5ab9L64-L78) [[2]](diffhunk://#diff-3e93cc11b22eab442b7c9435196e3101e2ac6f286ee1b63c98f0de4099bc5ab9L228-R288) [[3]](diffhunk://#diff-3e93cc11b22eab442b7c9435196e3101e2ac6f286ee1b63c98f0de4099bc5ab9L255-R304)
* Updated `ConfigHandler` in `config/config.go` to use the new `ConfigFileName` for configuration file paths.

### `VPNClient` Interface and Implementation:
* Simplified the `VPNClient` constructor by removing the `Options` struct and using direct parameters (`dataDir`, `logDir`, etc.).
* Removed unused fields (`DeviceID`, `Locale`) from the `Options` struct in `client/client.go`.
* Added a new `ActiveServer` method to `vpnClient` to retrieve the currently connected server.

### Code Cleanup and Deprecation:
* Removed the `OnNewConfig` method from both `vpnClient` and `BoxService` as it is no longer needed with the new configuration handling. [[1]](diffhunk://#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09L51) [[2]](diffhunk://#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09L262-L265) [[3]](diffhunk://#diff-3e93cc11b22eab442b7c9435196e3101e2ac6f286ee1b63c98f0de4099bc5ab9L228-R288)
* Deleted the `cmd/main.go` file, which contained a standalone entry point, as it is no longer relevant.

### Test Updates:
* Replaced the `TestOnNewConfig` test with `TestReloadOptions` to validate the new configuration reloading mechanism.

closes getlantern/engineering/issues/2121


garmr-ulfr:
@jigar-f I added you as a reviewer so you can confirm this will work on iOS. I made a couple of very minor changes to the radiance API itself, changed the args to `NewRadiance`, pulled the `APIHandler` into `Radiance` so they're initialized together. The big change though is that I pulled `VPNClient` out of `Radiance` so they can now be run independently of each other. Calling `Radiance.VPNClient` will initialize and return the VPN client. `VPNClient` has all of the functions/methods you are already using on `Radiance`, _e.g._ `StartVPN`, `StopVPN`, etc. 

Let me know if I need to make more changes for it to work. The intention is for everything that doesn't directly interact with the VPN client to be easily accessible from `Radiance` in the main process on iOS and Mac.

radiance.go
```diff
@@ -149,26 +198,12 @@ func (r *Radiance) Close() {
...
-		return nil, fmt.Errorf("no active server config")
-	}
-
-	return activeServer.(*Server), nil
-
+// VPNClient returns the [client.VPNClient] instance. The client will be initialized with the provided
+// [libbox.PlatformInterface] if has not been initialized yet, or will return the existing client instance.
+// platIfce used to interact with the underlying platform on iOS and Android. On other platforms, it
+// is ignored and can be nil. enableSplitTunnel is the initial state of split tunneling when the service starts.
+func (r *Radiance) VPNClient(platIfce libbox.PlatformInterface, enableSplitTunnel bool) (client.VPNClient, error) {
```
jigar-f, line 205: Slight problem here, can we make this method independent?

So the main process will start NewRadiance and do whatever needs to be done.

Now, coming to Tunnel, Tunnel will call NewRadiance-> VPNClient (Since VPNClient is connected to radiance instance), so NewRadiance will get called again in Tunnel (I am fine calling again if this operation is not heavy on I/O thread for mobile)

Can't we do something like this
```
// is ignored and can be nil. enableSplitTunnel is the initial state of split tunneling when the service starts.
func  VPNClient(dataDir,logDir string,platIfce libbox.PlatformInterface, enableSplitTunnel bool) (client.VPNClient, error) {
	return client.NewVPNClient(dataDir, logDir, platIfce, enableSplitTunnel)
}
```

jigar-f: cc @garmr-ulfr

internal/filewatcher.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"fmt"
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"golang.org/x/exp/slog"
```
WendelHime, line 11: Probably mean (will need to format imports)
```suggestion
	"log/slog"
```

WendelHime (approved):
LGTM!

internal/filewatcher.go
```diff
@@ -0,0 +1,104 @@
...
+		timer       *time.Timer
+		timerAccess sync.Mutex
+	)
+	for {
+		select {
+		case event, ok := <-fw.watcher.Events:
+			if !ok {
+				return
+			}
+			if event.Name != fw.path && !event.Has(fsnotify.Create) && !event.Has(fsnotify.Write) {
```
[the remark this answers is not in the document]
garmr-ulfr: We want to check the file at a specific path, not any file with the same name.

PR #97, opened 2025-05-21
feat: Lantern Server automatic provisioning with DO and GCP

reflog:
closes https://github.com/getlantern/engineering/issues/2118

This PR is "a bit large".
It includes implementation of all the APIs that are needed to provision a Lantern Server manager instance on user's behalf on DO or GCP.

It creates a VM, registers temporary ssh keys, prepares firewall, installs packages and reports the results back.
It handles OAuth flow for accessing the APIs on user's behalf.

I'd like @jigar-f @atavism to also take a look, since they'll have to integrate with this on the App side.
I'm thinking I might add a channel that will receive status updates like:
```go
{
 Message: "Installing packages",
 Attempt: 1,
 MaxTries: 10,
}
```

so that UI can update accordingly.

I've included `servers/cmd/main.go` that contains code that can be put into the client app and shows how to use the new functionality

Looking forward to hearing some feedback

### DIFF
--- a/servers/cloud/gcp/oauth.go
+++ b/servers/cloud/gcp/oauth.go
@@ -55,7 +55,7 @@ func RunOauth(openBrowser func(string) error) *common.OauthSession {
 }
 
 func startOauthFlow(ctx context.Context, openBrowser func(string) error, resultChan chan<- common.OauthResult) {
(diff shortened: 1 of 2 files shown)


garmr-ulfr (commented):
This feels like it's its own thing, like a completely separate tool that could be useful by itself, and probably should be in its own repo.

PR #116, opened 2025-07-23
Add options parameter to VPN connect functions

jigar-f:
This pull request introduces changes to improve the initialization and logging process in the `vpn/vpn.go` file. The most significant updates include adding support for passing configuration options via the `radiance.Options` struct, ensuring proper initialization of shared resources, and enhancing logging for debugging purposes.

### Initialization Improvements:
* Updated the `QuickConnect` and `ConnectToServer` functions to accept a new `radiance.Options` parameter, ensuring that paths and logger configurations are properly initialized using `common.Init`. This prevents potential issues with uninitialized resources. [[1]](diffhunk://#diff-25828e7146300d37842032752af02678c7c569a51995a9d0314e1ec400cdb94eR21-R42) [[2]](diffhunk://#diff-25828e7146300d37842032752af02678c7c569a51995a9d0314e1ec400cdb94eL51-R69)

### Logging Enhancements:
* Added a `log.Printf` statement in the `connect` function to log the data path, aiding in debugging and visibility of runtime behavior.

### Dependency Updates:
* Added `log` and `github.com/getlantern/radiance` to the import statements to support the new logging and options functionality. [[1]](diffhunk://#diff-25828e7146300d37842032752af02678c7c569a51995a9d0314e1ec400cdb94eR10) [[2]](diffhunk://#diff-25828e7146300d37842032752af02678c7c569a51995a9d0314e1ec400cdb94eR21-R42)

### DIFF
--- a/vpn/vpn.go
+++ b/vpn/vpn.go
@@ -17,7 +17,6 @@ import (
 	"github.com/sagernet/sing-box/experimental/cachefile"
 	"github.com/sagernet/sing-box/experimental/libbox"
 
-	"github.com/getlantern/radiance"
 	"github.com/getlantern/radiance/common"
 	"github.com/getlantern/radiance/servers"
 )
@@ -25,20 +24,12 @@ import (
 // QuickConnect automatically connects to the best available server in the specified group. Valid
 // groups are [ServerGroupLantern], [ServerGroupUser], "all", or the empty string. Using "all" or
 // the empty string will connect to the best available server across all groups.
-func QuickConnect(group string, platIfce libbox.PlatformInterface, options radiance.Options) error {
-	// Since this will start as new process we need to ask for path and logger.
-	// This is to ensure that the options are correctly set for the new process.
-	err := common.Init(options.DataDir, options.LogDir, options.LogLevel)
-	if err != nil {
-		// Better to return the error
-		// if this not working tunnel will not work.
-		return fmt.Errorf("failed to initialize common: %w", err)
-	}
+func QuickConnect(group string, platIfce libbox.PlatformInterface, dataDir, logDir, logLevel string) error {
 	switch group {
 	case servers.SGLantern:
-		return ConnectToServer(servers.SGLantern, autoLanternTag, platIfce, options)
+		return ConnectToServer(servers.SGLantern, autoLanternTag, platIfce, dataDir, logDir, logLevel)
 	case servers.SGUser:
-		return ConnectToServer(servers.SGUser, autoUserTag, platIfce, options)
+		return ConnectToServer(servers.SGUser, autoUserTag, platIfce, dataDir, logDir, logLevel)
 	case "all", "":
 		group = autoAllTag
 	default:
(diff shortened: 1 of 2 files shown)


vpn/vpn.go
```diff
@@ -17,19 +17,28 @@ import (
...
+	"github.com/getlantern/radiance"
 	"github.com/getlantern/radiance/common"
 	"github.com/getlantern/radiance/servers"
 )
 
 // QuickConnect automatically connects to the best available server in the specified group. Valid
 // groups are [ServerGroupLantern], [ServerGroupUser], "all", or the empty string. Using "all" or
 // the empty string will connect to the best available server across all groups.
-func QuickConnect(group string, platIfce libbox.PlatformInterface) error {
+func QuickConnect(group string, platIfce libbox.PlatformInterface, options radiance.Options) error {
```
garmr-ulfr, line 28: We should avoid import `radiance` to ensure that they remain completely independent. 
```suggestion
func QuickConnect(group string, platIfce libbox.PlatformInterface, dataDir, logDir, logLevel string) error {
```

vpn/vpn.go
```diff
@@ -48,7 +57,15 @@ func QuickConnect(group string, platIfce libbox.PlatformInterface) error {
 
 // ConnectToServer connects to a specific server identified by the group and tag. Valid groups are
 // [servers.SGLantern] and [servers.SGUser].
-func ConnectToServer(group, tag string, platIfce libbox.PlatformInterface) error {
+func ConnectToServer(group, tag string, platIfce libbox.PlatformInterface, options radiance.Options) error {
```
garmr-ulfr, line 60: Same as above.
```suggestion
func ConnectToServer(group, tag string, platIfce libbox.PlatformInterface, dataDir, logDir, logLevel string) error {
```

vpn/vpn.go
```diff
@@ -17,19 +17,28 @@ import (
...
 )
 
 // QuickConnect automatically connects to the best available server in the specified group. Valid
 // groups are [ServerGroupLantern], [ServerGroupUser], "all", or the empty string. Using "all" or
 // the empty string will connect to the best available server across all groups.
-func QuickConnect(group string, platIfce libbox.PlatformInterface) error {
+func QuickConnect(group string, platIfce libbox.PlatformInterface, options radiance.Options) error {
+	// Since this will start as new process we need to ask for path and logger.
+	// This is to ensure that the options are correctly set for the new process.
+	err := common.Init(options.DataDir, options.LogDir, options.LogLevel)
```
garmr-ulfr, line 31: Nit: I agree with Copilot that this is redundant here. We could move it down to line 55, right before `return connect(group, "", platIfce)`

jigar-f:
>   we should probably also allow them to be updated, rather than only setting them on the first call.

Why?, For macOS and iOS it will start in a new process, so there is no need to update it will live only when the tunnel is active.

vpn/vpn.go
jigar-f: Done.

jigar-f:
@garmr-ulfr all comment addressed

garmr-ulfr:
> Why?, For macOS and iOS it will start in a new process, so there is no need to update it will live only when the tunnel is active.

I was thinking more about when changing servers while still connected or usage on other platforms.

Requiring the parameters on every call feels a bit unexpected outside of iOS/macOS. Like calling `Log("message", "a.log")` then `Log("error", "e.log")` and not seeing both files with the correct log. 

Since this behavior is specific to iOS and macOS, it's common to have platform-specific APIs with separate functions for switching servers, or just a separate function to initialize the app/process.

Anyway, it's a minor thing, just sharing my thoughts :smiley:. This is fine!

jigar-f:
Thanks for the review.

PR #118, opened 2025-08-06
fix(boxoptions): unmarshal as config.Config

garmr-ulfr:
This pull request includes a couple of bug fixes. 

- `boxoptions.go`: loads config as `config.Config` instead of `LC.ConfigResponse`.
- `tunnel.go`: adds explicit `log.Factory` type parameter when adding it to the `ctx`. Both sing-box and lantern logs will now go to the same file.


garmr-ulfr:
Merging. Minor change.

PR #130, opened 2025-09-06
feat(servers/vpn): much more logs

garmr-ulfr:
Logs..just so much more logs.


garmr-ulfr:
Just adding more logs. Merging.

PR #159, opened 2025-10-03
feat(docs): update env variable instructions

garmr-ulfr:
This pull request updates the documentation for environment variable configuration in `README.md` to clarify where `.env` files can be placed and adds a new environment variable for log output control.

Documentation improvements:

* Clarified that configuration can be controlled by setting environment variables or creating a `.env` file in the same directory as the compiled executable or project root directory, rather than only in the project root.
* Updated the list of available environment variables to include `RADIANCE_DISABLE_STDOUT_LOG`, which disables printing `radiance` logs to STDOUT while still writing logs to the log file.


garmr-ulfr:
Just small update to README. Merging.

PR #408, opened 2026-04-10
fix: add panic recovery to IPC outbound handlers + increase timeouts

myleshorton:
## Summary
- Add `defer recover()` to `updateOutboundsHandler`, `addOutboundsHandler`, and `removeOutboundsHandler` — matching the pattern already used by `selectHandler`
- Increase IPC HTTP server timeouts from 5s read / 5s write to 15s read / 30s write to accommodate processing 30+ outbounds on config refresh

Without panic recovery, when `tunnel.updateOutbounds()` panics (e.g., missing selector/urltest group at `tunnel.go:455`), the HTTP connection drops silently with EOF. The caller sees `Failed to forward event to tunnel` every config refresh cycle (~3 min), meaning bandit URL overrides never reach the running tunnel.

## Test plan
- [ ] Verify `outbound/update` EOF errors stop after this fix
- [ ] Verify config refreshes successfully update outbounds in the running tunnel
- [ ] If a panic still occurs, it now returns HTTP 500 with the panic message instead of EOF

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/getlantern/fronted v0.0.0-20260325003030-cb5041ba1538
 	github.com/getlantern/keepcurrent v0.0.0-20260304213122-017d542145ae
 	github.com/getlantern/kindling v0.0.0-20260329144042-b1825b9cb1bb
-	github.com/getlantern/lantern-box v0.0.61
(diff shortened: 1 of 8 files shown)


myleshorton:
Superseded by a clean PR off main.

PR #409, opened 2026-04-10
fix: add panic recovery to IPC outbound handlers + increase timeouts

myleshorton:
## Summary

Process IPC outbound update/add/remove requests asynchronously. The handlers now return `202 Accepted` immediately and do the actual outbound processing in a background goroutine with panic recovery.

### Root cause

Every config refresh (~3 min), the `updateOutboundsHandler` processes 30+ outbounds — adding new ones, removing old ones — which takes ~12 seconds. The HTTP server's `WriteTimeout` is 5 seconds. After 5s the server kills the connection, causing the client to receive EOF. The handler keeps running to completion (outbounds are updated successfully), but the client never gets the response.

### Fix

Return immediately with `202 Accepted` after parsing/validating the request body. The actual outbound processing runs in a goroutine. Errors are logged server-side. The client accepts both `200 OK` and `202 Accepted`.

## Test plan
- [ ] Verify `outbound/update` EOF errors stop
- [ ] Verify outbounds are still updated correctly (server-side logs show processing)

### DIFF
--- a/vpn/ipc/http.go
+++ b/vpn/ipc/http.go
@@ -56,7 +56,7 @@ func sendRequest[T any](ctx context.Context, method, endpoint string, data any) 
 		return res, traces.RecordError(ctx, fmt.Errorf("request failed: %w", err))
 	}
 	defer resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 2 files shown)


myleshorton:
Replacing with a clean PR.

PR #414, opened 2026-04-14
Store the user ID and token for device limit flow

jigar-f:
This pull request updates the logic for handling login responses when legacy user data is missing. Now, if `LegacyUserData` is absent, the code will still store the user ID and token in the settings, and log more informative messages.

User data handling improvements:

* When `LegacyUserData` is missing in the login response, the code now attempts to store `LegacyID` and `LegacyToken` in the settings, and logs errors if these operations fail. The log message is also updated to clarify what is being stored.

### DIFF
--- a/api/user.go
+++ b/api/user.go
@@ -780,15 +780,17 @@ func (a *APIClient) setData(data *protos.LoginResponse) {
 			}
 		}
 
-		devices := []settings.Device{}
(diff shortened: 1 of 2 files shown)


jigar-f:
@WendelHime @garmr-ulfr Small quick fix, Pulling this in.

PR #422, opened 2026-04-16
ipc: add /config/events SSE endpoint for config-change notifications

myleshorton:
## Summary

On the refactor branch \`config.NewConfigEvent\` is emitted inside the packet-tunnel extension's radiance process. Subscribers in the host app — notably lantern-core's \`listenConfigEvents\`, which forwards these to Flutter as \`\"config\"\` events driving \`availableServersProvider.forceFetchAvailableServers()\` and \`homeProvider.fetchUserDataIfNeeded()\` — never see them because \`events.Subscribe\` is an in-memory fan-out that doesn't cross the process boundary.

This PR adds \`/config/events\` alongside \`/server/auto-selected/events\` and \`/vpn/status/events\`. Mirrors the existing pattern exactly.

The payload is intentionally empty (\`{}\`) — callers only need to know a change occurred and can fetch fresh state through the other GET endpoints. Streaming the full \`config.NewConfigEvent\` (which holds the entire old + new \`*Config\`) would waste bandwidth for no consumer benefit.

## Client
\`\`\`go
func (c *Client) ConfigEvents(ctx context.Context, handler func()) error
\`\`\`

Handler fires once per frame until ctx is cancelled. Same reconnect semantics as \`AutoSelectedEvents\` / \`DataCapStream\`.

## Companion PR

This is the radiance half. The lantern-core half adds \`go lc.listenConfigEvents()\` back in, now using this SSE stream instead of the in-process \`events.SubscribeContext\` that Patrick removed in \`4d4e06d9d\` (because it was dead on the refactor split).

Tracked overall in getlantern/engineering#3182.

## Test plan
- [x] \`go build ./ipc/...\`
- [x] \`go vet ./ipc/...\`
- [x] \`go test -count=1 ./ipc/...\`
- [ ] Smoke test: Flutter UI refreshes available servers/user data when a new config arrives

### DIFF
--- a/ipc/client.go
+++ b/ipc/client.go
@@ -193,6 +193,21 @@ func (c *Client) AutoSelectedEvents(ctx context.Context, handler func(vpn.AutoSe
 }
 
 ///////////////////////
+// Config events     //
+///////////////////////
+
+// ConfigEvents connects to the config event stream. The server emits a frame
+// on every config.NewConfigEvent; the payload is intentionally empty — callers
+// should treat each frame as a "refresh" signal and fetch any state they need
+// via the other GET endpoints. The handler is called once per frame received
+// until ctx is cancelled or the connection is closed.
+func (c *Client) ConfigEvents(ctx context.Context, handler func()) error {
+	return c.sseStream(ctx, configEventsEndpoint, func(data []byte) {
+		handler()
+	})
+}
+
+///////////////////////
 // Server management //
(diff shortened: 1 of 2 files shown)


garmr-ulfr (approved):
LGTM

garmr-ulfr:
I'm going to go ahead and bring this in.

PR #424, opened 2026-04-17
vpn: pre-download remote rule-sets before sing-box init

myleshorton:
## Summary

On Android, sing-box initializes remote rule-sets (\`geosite-cn\`, smart-routing categories like \`ai-domains\`) *before* the TUN is up. The DNS router has no registered network interface at that point, so every hostname lookup fails with \`exchange4: no available network interface | exchange6: no available network interface\` — including lookups routed to \`dns_local\` (a DoH server reached through sing-box's own dialer). The download blocks libbox startup, so the tunnel never comes up.

This PR fetches remote rule-sets in radiance, *before* handing options to libbox, using \`net/http\` (OS resolver + platform default dialer, both of which work pre-TUN). Each entry is rewritten to \`type=local\` pointing at a cached file under \`{dataPath}/rulesets/\`.

Behavior:
- First start: download via Go stdlib, cache on disk, start sing-box with local paths.
- Cached: skip HTTP if the cache is fresher than \`RemoteRuleSet.UpdateInterval\` (default 24h).
- Offline w/ stale cache: fall back to the stale cache (better than failing startup).
- Offline, no cache: leave as \`remote\` — same failure as today.

No platform-specific branching; every platform benefits from the caching.

## Context

Reproduced and verified on the Android emulator with \`RADIANCE_COUNTRY=MO\` (smart-routing country). Before: libbox fails with \`initial rule-set ... no available network interface\`. After: \`Pre-downloaded rule-set\` ×N → \`Tunnel initializated\` → VPN Status: **Connected**.

Original diagnosis: Freshdesk [#172564](https://lantern.freshdesk.com/a/tickets/172564) (Macao, CN Unicom — geosite-cn bootstrap deadlock) and [#172722](https://lantern.freshdesk.com/a/tickets/172722) (Bulgaria — smart-routing-bg bootstrap deadlock).

This **supersedes** the server-side \`dns_ruleset_host_bypass\` DNS rule in lantern-cloud [#2588](https://github.com/getlantern/lantern-cloud/pull/2588). That fix routed the rule-set hostnames to \`dns_local\`, but \`dns_local\` itself fails for the same reason — sing-box's dialer has no interfaces at init regardless of which DNS server the route chooses. The feature flag can be retired after this lands.

Companion PR against \`refactor\`: #423.

## Test plan

- [x] \`go test ./vpn/... -run TestPreDownloadRuleSets\` — 4 new unit tests (rewrite, cache hit, stale fallback, local passthrough)
- [x] Android emulator with \`RADIANCE_COUNTRY=MO\` — tunnel connects cleanly
- [ ] Desktop regression — rule-sets still update normally (UpdateInterval=24h)
- [ ] iOS — same bootstrap issue theoretically exists; verify no regression

### DIFF
--- a/vpn/rulesets.go
+++ b/vpn/rulesets.go
@@ -1,0 +1,129 @@
+package vpn
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+
+	C "github.com/sagernet/sing-box/constant"
+	O "github.com/sagernet/sing-box/option"
+
+	"github.com/getlantern/radiance/common/atomicfile"
+)
+
+const (
+	ruleSetsCacheDir     = "rulesets"
+	ruleSetDownloadLimit = 32 << 20 // 32 MiB — upper bound on a single rule-set file
+	ruleSetDownloadTotal = 60 * time.Second
+	defaultRuleSetTTL    = 24 * time.Hour
+)
+
+// preDownloadRuleSets rewrites every remote rule-set in opts to a local one,
+// downloading content via Go's standard net/http (OS resolver, default dialer)
+// into dataPath/rulesets. This sidesteps a bootstrap deadlock on Android where
+// sing-box's DNS/dialers have no registered network interface at rule-set init
+// time, so every hostname lookup fails with "no available network interface"
+// — regardless of which DNS server the route chooses.
+//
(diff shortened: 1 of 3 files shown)


myleshorton:
Closing — same reasoning as #423: the real fix lives in sing-box-minimal commit 9c79c311 (v1.12.21-lantern), which makes initial remote rule-set fetch non-fatal. Once lantern main's go.mod is bumped back to v1.12.21-lantern (it was accidentally downgraded in lantern#8655), the bootstrap deadlock goes away without any radiance-level pre-download.

PR #425, opened 2026-04-17
vpn: integrate unbounded outbound via direct-transport context

myleshorton:
## Summary

Integrates lantern-box's new `unbounded` outbound (lantern-box#241) into radiance so censored clients can use peer-matched WebRTC-egress tunnels alongside the existing outbounds.

Two changes:

- **`vpn/tunnel.go`** — wraps `box.BaseContext()` with `adapter.ContextWithDirectTransport(ctx, kindling.HTTPClient().Transport)` before constructing the libbox service. The unbounded outbound reads this value to dial freddie for signaling; without it, signaling traffic re-enters the VPN tunnel and loops.
- **`go.mod`** — bumps `github.com/getlantern/lantern-box` to a pseudo-version of the `fisk/unbounded-outbound-v2` branch so the compile-time symbols (`adapter.ContextWithDirectTransport`, `protocol.SupportedProtocols()` entry for `"unbounded"`) resolve.

No other radiance code changes are needed:

- `box.BaseContext()` already calls `protocol.RegisterProtocols(ctx)`, so the unbounded outbound is registered with the sing-box registry automatically.
- `config/fetcher.go:86` already sends `protocol.SupportedProtocols()` in the config request, so the server will now know the client supports `"unbounded"` and can include it in outbound assignments without any server-awareness code here.

## Status

**Draft, blocked on lantern-box#241.** Merging depends on:
1. lantern-box#241 landing on `main`
2. A new lantern-box tag (e.g. v0.0.69)
3. A follow-up commit here replacing the pseudo-version with the tag and re-running `go mod tidy`

## Test plan

- [ ] CI passes on the current pseudo-version pin
- [ ] Once lantern-box#241 merges, replace pin with real tag and confirm build + `go test ./vpn/...` still pass
- [ ] Manual: build Android APK from this branch, enable a server config with an `"unbounded"` outbound, confirm the client dials freddie via the physical interface (not the tunnel) and establishes a session
- [ ] Confirm no regressions on non-unbounded outbounds (the direct-transport context value is unused by other outbounds)


myleshorton:
Superseded by #429 — the direct-transport + streaming-wrapper for Unbounded signaling now lives there using kindling's HTTPClient().Transport (no AMP via Accept: text/event-stream). This draft's freddieTransport hardcoded 1.1.1.1 as a DNS resolver, which is typically blocked in censored countries and would defeat the whole censorship-circumvention point. All useful content from here is in #429 without that regression.

PR #427, opened 2026-04-19
env: don't let SetStagingEnv override a shell-set RADIANCE_ENV

myleshorton:
## Summary

\`SetStagingEnv\`'s docstring says \"sets the environment to staging if it has not already been set\" but the implementation unconditionally overwrote \`envVars[ENV]\`. Honor the docstring.

## Why

The Flutter UI persists its current environment in app settings and passes it as \`Opts.Env\` on every launch. When that persisted setting is \"staging\", lantern-core calls \`SetStagingEnv\` and clobbers whatever the init loop already populated from the shell \`RADIANCE_ENV\` export.

Observed today on a staging unbounded end-to-end test: developer runs

\`\`\`
RADIANCE_ENV=staging RADIANCE_FEATURE_OVERRIDES=bandit_proxy_assignment \\
  /Applications/Lantern.app/Contents/MacOS/Lantern
\`\`\`

— wants staging — but the Flutter UI's persisted setting (from some earlier GUI session) interferes. The mirror-image bug on this main branch is equally silent: shell says \`prod\`, Flutter persisted \`staging\`, SetStagingEnv wipes \`prod\` → developer is unexpectedly pointed at staging.

The fix preserves the dev-ergonomics side effect (\`envVars[PrintCurl] = true\`) so staging-on-request instrumentation still works, and a new test rejects any future regression to unconditional overwrite.

## Test plan
- [x] \`go test ./common/env/...\` — both new tests pass
- [x] \`go build ./...\` clean

### DIFF
--- a/common/env/env.go
+++ b/common/env/env.go
@@ -93,9 +93,19 @@ func Get[T any](key Key) (T, bool) {
 	return zero, false
 }
 
-// SetStagingEnv sets the environment to staging if it has not already been set.
-// This is used for testing that need to interact with staging services,
+// SetStagingEnv sets the environment to staging if it has not already
+// been set. Callers typically invoke this when the Flutter UI's
+// persisted `environment` setting is "staging", but that persisted
+// setting must not override a developer's explicit shell env —
+// RADIANCE_ENV from the shell wins. If RADIANCE_ENV is already set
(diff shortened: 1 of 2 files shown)


myleshorton:
Retargeting against refactor branch — the actual codebase variant with the reversed-precedence bug.

PR #428, opened 2026-04-19
env: flip Get() precedence — OS env wins over dotenv, matching docstring

myleshorton:
## Summary

Package docstring says \"environment variables > configurations set at .env file\" but \`Get()\` checked dotenv FIRST and only fell back to \`os.LookupEnv\` when the key was missing.

## Why this bit us today

The Flutter UI persists its current environment as an app-setting (e.g. \`environment: prod\`) and passes it into lantern-core on every launch. When that persisted setting is \`\"staging\"\`, core calls \`SetStagingEnv()\` which writes \`dotenv[RADIANCE_ENV]=\"staging\"\`. Under the old precedence, a developer starting the app with

\`\`\`
RADIANCE_ENV=prod /Applications/Lantern.app/Contents/MacOS/Lantern
\`\`\`

silently got flipped to staging — frontend clobbering CLI override. Observed today while trying to pin a desktop client at staging for Unbounded end-to-end verification.

## Change

Flip to OS env first, dotenv fallback.

- Runtime \`Set\` / \`.env\` values are still consulted for keys the shell hasn't explicitly set, so \`SetStagingEnv\` and the \`/env\` IPC endpoint keep working for users who don't export anything themselves.
- Updated \`Set()\`'s docstring — the claim that \`Set\` \"overrides OS env\" is no longer true.

## Tests

- \`TestGet_OSEnvWinsOverDotenv\` — both set, OS must win.
- \`TestGet_DotenvFallsBackWhenOSUnset\` — only dotenv set, Get must still surface it.

Both fail under the old precedence.

## Test plan
- [x] \`go build ./common/env/... ./backend/... ./config/... ./ipc/...\` clean
- [x] \`go test ./common/env/... -count=1 -v\` — new tests pass
- [ ] \`cmd/lantern\` has a pre-existing build break on this branch (ipc.NewClient signature drift), not caused by this PR

### DIFF
--- a/common/env/env.go
+++ b/common/env/env.go
@@ -64,21 +64,34 @@ func init() {
 	}
 }
 
+// Get resolves a key with OS environment variables taking precedence
+// over values loaded from a .env file or written at runtime via Set.
+// This matches the package-level docstring ("environment variables >
+// configurations set at .env file") and lets a developer's explicit
+// shell export (e.g. `RADIANCE_ENV=staging Lantern`) win over whatever
+// the Flutter UI's persisted app-settings ends up passing through
+// SetStagingEnv or the /env IPC endpoint — otherwise a stale persisted
+// setting silently overrides the command-line intent.
+//
+// The dotenv map is still consulted as a fallback so runtime overrides
+// (from .env file or IPC) take effect for keys the shell hasn't set.
 func Get(key _key) (string, bool) {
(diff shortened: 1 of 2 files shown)


myleshorton:
Superseded by #429 which combines this precedence fix with the `Options.EnvOverrides` plumbing needed to propagate shell env into the sandboxed system extension.

PR #434, opened 2026-04-22
issue-report: add multipart attachment support

atavism:
Depends on https://github.com/getlantern/lantern-cloud/pull/2628

For https://github.com/getlantern/engineering/issues/3164

Send screenshot attachments to Freshdesk as first-class multipart files while preserving the legacy protobuf path for reports without screenshots


issue/issue.go
```diff
@@ -41,6 +41,13 @@ func NewIssueReporter(httpClient *http.Client) *IssueReporter {
 
 type IssueType int
 
+type Attachment struct {
+	Name       string
+	Type       string
+	Data       []byte
+	FirstClass bool
```
garmr-ulfr: Do we need this? Since we're just allowing screenshots I think we can get rid of this and just treat what's provided as first class.

issue/transport.go
```diff
@@ -0,0 +1,219 @@
...
+	name := strings.TrimSpace(attachment.Name)
+	if contentType := normalizeAttachmentType(
+		mime.TypeByExtension(strings.ToLower(filepath.Ext(name))),
+	); contentType != "" {
+		return contentType
+	}
+
+	if len(attachment.Data) == 0 {
+		return octetStreamContentType
+	}
```
garmr-ulfr: We could move this to the top with the `attachment == nil` check since there's no reason to try to detect the type if there's no data.

issue/transport.go
```diff
@@ -0,0 +1,219 @@
...
+			continue
+		}
+
+		name, err := normalizeAttachmentName(attachment.Name)
+		if err != nil {
+			return err
+		}
+		if len(attachment.Data) == 0 {
+			return fmt.Errorf("attachment %q is empty", name)
+		}
```
garmr-ulfr: Same here.

issue/issue.go
```diff
@@ -80,6 +87,9 @@ type IssueReport struct {
 	Locale            string
 	// device alphanumeric name
 	Model string
+	// Attachments contains in-memory attachments supplied by the caller. FirstClass
+	// attachments are sent as separate multipart files.
```
garmr-ulfr: We should include the number and size limits so it's in the doc displayed by IDEs.  We could use `[maxFirstClassAttachmentCount]` and `[maxFirstClassAttachmentBytes]` so the comment doesn't become outdated if the values change.

ipc/client.go
```diff
@@ -628,9 +628,15 @@ func (c *Client) VerifySubscription(ctx context.Context, service account.Subscri
 // ReportIssue submits an issue report. additionalAttachments is a list of file paths for additional
 // files to include. Logs, diagnostics, and the config response are included automatically and do
 // not need to be specified.
```
garmr-ulfr: Same here about limit restrictions. We'd need to export the consts though.

issue/transport.go
```diff
@@ -0,0 +1,219 @@
...
+	"mime/multipart"
+	"net/http"
+	"net/textproto"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	maxFirstClassAttachmentCount = 3
+	maxIssueAttachmentBytes      = 15 * 1024 * 1024
```
garmr-ulfr: Can we rename this to `maxFirstClassAttachmentBytes` so it's not confused as the overall total, including logs?

garmr-ulfr (commented):
Could we change naming to screenshots instead of first class? It makes it obvious that that it's only for screenshots and not anything else.

garmr-ulfr (approved):
LGTM!

atavism:
Thanks for the review, @garmr-ulfr!

PR #438, opened 2026-04-22
vpn: route remote rule-set downloads through ipv4-only detour

myleshorton:
Fixes 'connected but no internet' on Russian cellular users whose direct path has no working IPv6 (Freshdesk #173534).

## Problem

With the default `direct` outbound, rule-set fetches to `raw.githubusercontent.com` fail with "network is unreachable" on Megafon and similar networks. The geosite/geoip bypass rules never load, `strict_route:true` sends Russian domains through the failing proxy chain, and the user sees a connected VPN with no working traffic.

## Fix

- New `direct-ipv4` outbound with `DomainStrategy: ipv4_only`
- `redirectRuleSetDetourToIPv4()` rewrites any remote rule-set whose `DownloadDetour` is empty or `direct` to use `direct-ipv4`

Scope is intentionally narrow: only the startup rule-set fetch path, not user traffic.

## Test plan

- [x] `go vet ./vpn/...` clean
- [ ] On a Russian network (or any IPv6-blocking cellular link), verify rule-set downloads succeed and geosite-ru / geoip-ru rules activate

### DIFF
--- a/vpn/boxoptions.go
+++ b/vpn/boxoptions.go
@@ -116,6 +116,25 @@ func baseOpts(basePath string) O.Options {
 				Tag:     "direct",
 				Options: &O.DirectOutboundOptions{},
 			},
+			// direct-ipv4 is used by remote rule-set downloads (see
+			// redirectRuleSetDetourToIPv4). Russian/cellular users frequently
+			// have no working IPv6 path to raw.githubusercontent.com; with a
+			// vanilla direct outbound the fetch returns "network is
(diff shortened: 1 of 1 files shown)


myleshorton:
Superseded — moving this fix to lantern-cloud so it reaches existing beta users via `/config-new` without an app rebuild. Server will emit both the `direct-ipv4` outbound and set `DownloadDetour: direct-ipv4` on rule-sets.

PR #445, opened 2026-04-27
qa: route all radiance egress through an upstream SOCKS5 + bandit probe driver

myleshorton:
Adds a development/QA-only path for routing **every** outbound network call radiance makes through a configured upstream SOCKS5. Pair with `pinger bridge --country ru` (lantern-cloud PR) to exercise the bandit + tunnel as if from a Russia residential client — no Android emulator required for the bandit-correctness test, but also unlocks the same path inside the Android client (lantern PR).

## Summary

- **`common/env`**: new key `RADIANCE_OUTBOUND_SOCKS_ADDRESS`, distinct from the existing `RADIANCE_SOCKS_ADDRESS` (which sets up an *inbound* listener — different concept).
- **`kindling/client`**: when the env var is set, swap kindling's stacked transport for a plain `http.Transport` whose `DialContext` dials through the upstream SOCKS5. Kindling's fronted/AMP/dnstt/proxyless circumvention paths are bypassed (kindling has no generic dialer override) — tracked as a follow-up if we want full fidelity.
- **`vpn/boxoptions`**: append a synthetic SOCKS5 outbound (`_dev_outbound_socks`) and walk every leaf outbound, setting `DialerOptions.Detour` to it. Selector / urltest / block / dns / **direct** are skipped (`direct` rejects Detour at runtime: \`detour is not supported in direct context\`).
- **`bypass/bypass`**: route through the same SOCKS5 instead of the local-bypass-proxy / direct-fallback chain.
- **`telemetry/otel`**: skip OTLP gRPC init in this mode — the gRPC exporter can't be routed via `http.Transport.DialContext`, so leaving it on would leak the test process's real IP.
- **`backend/radiance`**: skip `publicip.Detect()` when the env var is set. Without this, radiance would talk directly to AWS/ifconfig.me, get the host's real IP, and stuff it into `X-Lantern-Config-Client-IP` — which the API trusts over the actual TCP source for the bandit lookup.
- **`config/fetcher`**: when the env var is set, drop UDP-only protocols (hysteria/hysteria2/wireguard/tuic/amnezia) from the request's `Protocols` list. The bridge SOCKS5 listener on the lantern-cloud side is TCP-only; without this filter, the bandit assigns UDP outbounds that fail with SOCKS5 \`code=7\` (Command not supported) until URLTest converges. Hysteria-class protocols don't work in Russia today anyway.
- **`common/platform`**: change `Platform` from `const` to `var`; honor `RADIANCE_PLATFORM` env override in `common.Init()`. Lets a Go process running on macOS impersonate Android for the bandit's view of the client.
- **`cmd/qa-bandit`**: focused QA driver that boots a radiance backend with all of the above wired, captures the first `/v1/config-new` response, dumps the bandit assignment, then \`ConnectVPN(AutoSelect)\` and probes a target URL through the local SOCKS5 inbound to verify both the API view and the actual tunnel egress IP.

## Non-obvious: TZ + locale also need to match the country

This caught us during testing and is worth surfacing — without these, the API still sees \`country=US\` even when the request egresses through a Russia residential IP, and the bandit serves US-tier outbounds.

\`cmd/api/maxmind.go:LookupCountryASNState\` overrides the GeoIP-derived country with the *timezone-derived* country (or \`X-Lantern-Locale\`-derived) when they disagree, on the assumption that a mismatch means the client is behind a VPN. Production behavior is correct — a user in Moscow whose VPN egresses through a US server should still be treated as Russian for content/track decisions. But for QA, our request egresses through Russia residential while \`time.Now().Location()\` on the dev box is \`America/Denver\` (or wherever), so the API sees:

- IP geo: Russia (from the residential pool)
- Timezone hint: \`America/Denver\` → US
- → "client is behind a VPN, prefer header-derived country" → returns \`country=US\` ❌

The fix in \`cmd/qa-bandit\` is two flags, both default to Russia:

\`\`\`
--tz Europe/Moscow      # exported as \`TZ\` env var before any time call;
                         # radiance picks it up via timezone.IANANameForTime
--locale ru_RU          # passed to backend.Options.Locale → X-Lantern-Locale header
\`\`\`

Pass matching values for any other country (\`--tz Asia/Tehran --locale fa_IR\` for Iran, etc.). The Android-side equivalent is \`adb shell setprop debug.lantern.tz Europe/Moscow\` plus the locale set in the Lantern app's settings.

## Quick start

Pair with the lantern-cloud bridge:

\`\`\`
# Terminal 1 — bridge (lantern-cloud)
mise r pinger:bridge

# Terminal 2 — radiance
RADIANCE_OUTBOUND_SOCKS_ADDRESS=127.0.0.1:1080 \\
  go run -tags 'with_quic,with_gvisor,with_wireguard,with_utls' ./cmd/qa-bandit
\`\`\`

## Verified behavior

\`\`\`
API saw client as     : country=RU ip=85.172.81.50      # Yaroslavl, Rostelecom AS12389
Outbounds (6): samizdat (5x) + reflex (1x), all in Russia-tier locations
probe OK in 1.30s — egress IP: 141.148.227.249           # NL Oracle (samizdat-NL)
\`\`\`

## Test plan

- [ ] \`go run -tags '...' ./cmd/qa-bandit\` against a running pinger bridge: confirm \`country=RU\` in the bandit assignment dump and a non-home egress IP from the probe.
- [ ] \`go run ... ./cmd/qa-bandit --tz America/Denver --locale en_US\` (TZ override only): confirm the API now reports \`country=US\` even with the Russia residential IP — verifies the MaxMind/timezone gotcha above.
- [ ] \`go test ./vpn/... ./kindling/... ./bypass/... ./common/...\` (existing tests still pass with the var/const change).
- [ ] Without the env var: existing behavior (kindling stacked transports, direct sing-box outbound dials, full publicip.Detect, telemetry on) — no regression.


teamlantern:
This PR has been inactive for 30 days.

Waiting on: none

It will be closed automatically in 7 days if no action is taken.

teamlantern:
This PR has been inactive for 30 days.

Waiting on: none

It will be closed automatically in 7 days if no action is taken.

PR #448, opened 2026-04-29
Revert signup url and guard check for the sign up request.

jigar-f:
This pull request updates the user sign-up logic in `account/user.go` to enforce that both a pro token and user ID are present before proceeding. It also ensures these values, along with the device ID, are sent as headers with the sign-up request, improving user account linking and error handling.

**Sign-up flow improvements:**

* The sign-up process now checks that both the pro token and user ID are set in `settings`; if either is missing, it returns an error and halts the operation.
* The sign-up request now includes the device ID, pro token, and user ID as headers, ensuring the backend can correctly link the sign-up to an existing user when appropriate.
* The code now uses `sendRequest` instead of `sendProRequest` to make the sign-up API call, reflecting the updated requirements for headers.

### DIFF
--- a/account/user.go
+++ b/account/user.go
@@ -156,7 +156,6 @@ func (a *Client) SignUp(ctx context.Context, email, password string) ([]byte, *p
 
 	proToken := settings.GetString(settings.TokenKey)
 	userID := settings.GetString(settings.UserIDKey)
-	slog.Debug("Signing up user", "email", email, "proTokenSet", proToken)
 	if proToken == "" || userID == "" {
 		return nil, nil, traces.RecordError(ctx, errors.New("signup requires the pro token and user ID to continue"))
 	}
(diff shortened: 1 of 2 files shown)


jigar-f:
@garmr-ulfr Pulling this in, Small changes.

account/user.go
```diff
@@ -153,9 +153,19 @@ func (a *Client) SignUp(ctx context.Context, email, password string) ([]byte, *p
...
+	if proToken == "" || userID == "" {
+		return nil, nil, traces.RecordError(ctx, errors.New("signup requires the pro token and user ID to continue"))
+	}
+	headers := map[string]string{
+		common.DeviceIDHeader: settings.GetString(settings.DeviceIDKey),
+		common.ProTokenHeader: proToken,
+		common.UserIDHeader:   userID,
+	}
+
+	resp, err := a.sendRequest(ctx, "POST", "/users/signup", nil, headers, data)
```
garmr-ulfr, line 168: @jigar-f, the send request line that was removed was to the Pro server, just wanted to make sure that you did actually mean for it to send to the non-Pro server. `sendRequest` vs `sendProRequest`.

jigar-f: Yup, that's correct. I thought I could use pro request, but that chanegs base URL.

PR #452, opened 2026-04-30
Honor dev country override in config requests

atavism:
For https://github.com/getlantern/engineering/issues/3202


garmr-ulfr:
@atavism radiance already allows country to be overridden. Is it not sticking???

PR #461, opened 2026-05-04
fix: disconnect VPN on backend close

garmr-ulfr:
Disconnect the VPN in `LocalBackend.Close` to prevent network from entering an error state if VPN was connected.


garmr-ulfr:
Bringing this in, just adds an explicit call to disconnect VPN on close

PR #464, opened 2026-05-06
settings: also migrate from pre-9.x flashlight/lantern-client YAML

myleshorton:
## Summary

Follow-up to #463. That PR covered the v9.0.x → v9.1.x → fixed upgrade path. This one extends the same migration to recover state from the **pre-9.x flashlight + lantern-client** clients on desktop and iOS — anyone upgrading from v8.x (or anything with that schema) directly to v9.1+ was hitting the same "Pro lost on upgrade" symptom Derek reproduced for the v9.0.x case, because their good state lives in a YAML file the radiance migration didn't know about.

## What's added

A fourth candidate path in `migrateLegacySettingsIfNeeded`. Paths match exactly what `getlantern/appdir.General("Lantern")` produced in the pre-9.x desktop clients (verified against `lantern-client/desktop/app/settings.go:132` and `appdir/appdir_*.go`):

| Platform | Pre-9.x file |
|----------|-------------|
| **macOS**    | `~/Library/Application Support/Lantern/settings.yaml` |
| **Windows**  | `%APPDATA%\Lantern\settings.yaml` |
| **Linux**    | `~/.config/lantern/settings.yaml`  *(lowercase — `appdir_linux.go` lowercases the app name)* |
| **iOS**      | `<fileDir>/userconfig.yaml` *(sandbox-relative)* |

All three desktop platforms resolve via `os.UserConfigDir()` (just like the pre-9.x client did), so the lookup respects `$XDG_CONFIG_HOME` on Linux and the standard config-dir conventions on macOS/Windows.

Each candidate is read, parsed (YAML), and translated into the canonical `settings.json` schema before being handed to the same priority-pick logic the existing JSON candidates use.

### Priority order
1. canonical `settings.json`
2. v9.0.x `local.json`
3. **pre-9.x platform-specific YAML** ← new
4. v9.1.x `data/settings.json` (known-bugged)

The pre-9.x YAML beats the v9.1.x nested file because the nested file is known to have a bogus device id and possibly wrong user id (that was the whole point of #463), while the YAML is the user's real legitimate state from the prior client.

### Field translations

**Desktop** (the older flashlight client wrote bool `userPro`):
- `userID` → `user_id`
- `deviceID` → `device_id`
- `userPro` (bool) → `user_level` (`"pro"` if true; `"free"` if user_id is set; unset otherwise)
- `userToken` → `token`
- `emailAddress` → `email`

**iOS** (didn't persist user level locally):
- `UserID` → `user_id`
- `DeviceID` → `device_id`
- `Token` → `token`
- (no user_level — let next \`/account/login\` decide)

## What's NOT in this PR

**Android pre-9.x state lives in an encrypted SQLite** (\`/data/data/org.getlantern.lantern/files/masterDBv2/db\`, decrypted with a password from \`EncryptedSharedPreferences\`). Reading that requires Kotlin-side code in the lantern repo, not a pure-Go migration in radiance. That'll be a separate change. Anyone who upgraded through v9.0.x first already has their state migrated to \`local.json\`, which #463 handles; only direct-from-v8.x Android upgraders are affected, which is a smaller cohort.

## Test plan
- [x] \`go test ./common/settings/ -count=1\` — passes (12 sub-tests covering both new YAML migration paths and existing JSON candidate logic)
- [x] Test redirects \`legacyYAMLPathFn\` so the test temp dir is used in place of the host's actual \`~/Library/Application Support/Lantern/settings.yaml\` (otherwise the host's real Lantern install would interfere with the unit tests)
- [ ] Manual macOS upgrade test: install v8.x desktop → log in to Pro → upgrade to a build with this fix → verify Pro state survives via \`~/Library/Application Support/Lantern/settings.yaml\` migration
- [ ] Manual Linux upgrade test: same shape with \`~/.config/lantern/settings.yaml\`
- [ ] Manual Windows upgrade test: same shape with \`%APPDATA%\Lantern\settings.yaml\`
- [ ] Manual iOS upgrade test: same shape with sandbox-relative \`userconfig.yaml\`
- [ ] Confirm fresh installs still work (no migration trigger, no-op path)

### DIFF
--- a/common/settings/legacy_yaml.go
+++ b/common/settings/legacy_yaml.go
@@ -23,10 +23,14 @@ import (
 // readable from Go; Android stored its state in an encrypted SQLite
 // and needs a Kotlin-side migration):
 //
-//	macOS:   ~/.lantern/settings.yaml         (desktop schema)
-//	Windows: %APPDATA%\Lantern\settings.yaml  (desktop schema)
-//	Linux:   ~/.config/lantern/settings.yaml  (desktop schema)
-//	iOS:     <fileDir>/userconfig.yaml        (ios schema)
+//	macOS:   ~/Library/Application Support/Lantern/settings.yaml  (desktop)
+//	Windows: %APPDATA%\Lantern\settings.yaml                      (desktop)
+//	Linux:   ~/.config/lantern/settings.yaml                      (desktop)
+//	iOS:     <fileDir>/userconfig.yaml                            (ios)
+//
+// These match the path the pre-9.x flashlight + lantern-client used
+// (getlantern/appdir.General("Lantern"), with the linux build tag
+// lowercasing the app name).
 //
 // Field translation (desktop → canonical):
 //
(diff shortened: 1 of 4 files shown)


garmr-ulfr (approved):
LGTM. Claude, would you clean up the comments though?

PR #478, opened 2026-05-12
fix: setting water dir when adding or loading local config

WendelHime:
This PR uses [lantern-box](https://github.com/getlantern/lantern-box/pull/261) for fixing the VPN connection issues with a WATER outbound available. This PR also replaces the `water_dir` value for WATER outbounds so if a user adds a WATER outbound, radiance will replace with the correct value and use the provided dir to store the WASM files.

Part of getlantern/engineering#3095


backend/radiance.go
```diff
@@ -581,7 +581,10 @@ func (r *LocalBackend) setServers(list servers.ServerList, isLantern bool) error
 	if err := r.srvManager.SetServers(list, isLantern); err != nil {
 		return fmt.Errorf("failed to set servers in ServerManager: %w", err)
 	}
-	err := r.vpnClient.UpdateOutbounds(list)
+	// updateOutbounds evicts any outbound absent from the list; include all
+	// servers so user-added outbounds aren't removed on a Lantern config update.
+	allList := servers.ServerList{Servers: r.srvManager.AllServers(), URLOverrides: list.URLOverrides}
+	err := r.vpnClient.UpdateOutbounds(allList)
```
garmr-ulfr: @WendelHime this is actually addressing a pretty big bug. I'm pulling this into another [PR](https://github.com/getlantern/radiance/pull/482) that also touches this area.

Great find BTW!!

servers/manager.go
```diff
@@ -624,6 +628,56 @@ func (m *Manager) RevokePrivateServerInvite(ip string, port int, accessToken str
...
+// servers also use a writable directory instead of a caller-supplied value
+// (e.g. /tmp from a Linux-targeting config) that may not exist on the device.
+func (m *Manager) applyWATERDir(outbounds []option.Outbound) {
+	dir := m.waterDir()
+	for _, ob := range outbounds {
+		if opts, ok := ob.Options.(*lbO.WATEROutboundOptions); ok {
+			opts.Dir = dir
+		}
+	}
+}
```
garmr-ulfr: The WATER directory should be handled by the WATER outbound itself; it already has an option to provide the directory https://github.com/getlantern/lantern-box/blob/main/option/water.go#L32. This could just be set in `buildOptions`.

vpn/vpn.go
```diff
@@ -543,6 +545,17 @@ func (c *VPNClient) RunOfflineURLTests(basePath string, outbounds []option.Outbo
...
+	// Avoid mutating the caller's slice.
+	outbounds = append([]option.Outbound(nil), outbounds...)
+	for i := range outbounds {
+		// Seeding starts a long-lived BitTorrent client; disable it for this
+		// short-lived instance so instance.Close() returns promptly.
+		if waterOpts, ok := outbounds[i].Options.(*lbO.WATEROutboundOptions); outbounds[i].Type == lbC.TypeWATER && ok {
+			cp := *waterOpts
+			cp.SeedEnabled = false
+			outbounds[i].Options = &cp
+		}
```
garmr-ulfr, line 557: Good call!

servers/manager.go
WendelHime: Yeah but in this case the user might not know which directory to use when adding or setting the water_dir (the same applies to fetched configurations). Using a custom directory that doesn't exist or the application doesn't have permission to use would break the app. The goal here is to overwrite the directory with the one that radiance manages

WendelHime: hmmmm but yeah I guess we could also apply that on buildOptions

garmr-ulfr: Right, we would just set `WATEROutboundOptions.Dir` to the data directory in `buildOptions` so the correct directory is handed to the outbound.

WendelHime: fixed in [19bb5d7](https://github.com/getlantern/radiance/pull/478/commits/19bb5d78c40cb357770f08bd764d823f544e294b)

servers/manager_test.go
```diff
@@ -267,6 +267,7 @@ func TestSaveServersConcurrent(t *testing.T) {
 	}
 }
 
+
```
garmr-ulfr: Got a stray newline.

garmr-ulfr (approved):
LGTM! Just got a stray newline.

PR #481, opened 2026-05-14
peer: route the peer-box sing-box logger into radiance's slog

myleshorton:
## Summary

The second sing-box instance the peer client spins up was constructed with a plain \`box.BaseContext()\` and no sing-box log factory registered. That means router / dial / outbound errors from the peer-box landed on the box's default stderr logger and never reached \`lantern.log\` — which made the SmC verify-failure incident triaged today unnecessarily hard.

We had:
- \`[samizdat] CONNECT api.iantem.io:443: handler started\` at \`20:41:35.079\`
- \`[samizdat] CONNECT api.iantem.io:443: handler returned, starting drain\` at \`20:41:35.096\` — 17 ms later
- Then 5 s of drain timeout and 1475 bytes of the verifier's TLS ClientHello discarded
- Server-side: \`peer verify failed: verify round-trip through peer: Get \"https://api.iantem.io/v1/peer/verify-callback?token=...\": EOF\`

The peer's sing-box outbound dial to \`api.iantem.io\` failed almost instantly — but with **zero error log anywhere** in either \`lantern.log\` or \`lantern_macos.log\`. The \`i.logger.ErrorContext(ctx, err)\` at \`lantern-box/protocol/samizdat/inbound.go:178\` would have caught the underlying error, but the peer-box's logger output goes to a default stderr destination instead of slog.

## Change

Mirror what \`vpn/tunnel.go:141-142\` already does for the main tunnel's box — register \`lblog.NewFactory(slog.Default().Handler())\` as the \`sblog.Factory\` on the peer-box's context before \`NewServiceWithContext\`. sing-box logger output now flows through radiance's slog and into \`lantern.log\` alongside everything else.

## Test plan

- [x] \`go build ./peer/...\` passes
- [x] \`go test -count=1 ./peer/...\` passes (existing peer tests still green)
- [ ] Next SmC toggle that triggers a verify failure should now produce a sing-box-side error log line in \`lantern.log\` explaining the actual dial failure — that's the signal we were missing today.

## Stack note

Branched off \`fisk/peer-connection-events\` because the peer module lives there. Will rebase onto main once the peer module merges.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -13,8 +13,11 @@ import (
 	"time"
 
 	"github.com/sagernet/sing-box/experimental/libbox"
+	sblog "github.com/sagernet/sing-box/log"
+	"github.com/sagernet/sing/service"
 
 	box "github.com/getlantern/lantern-box"
+	lblog "github.com/getlantern/lantern-box/log"
 	"github.com/getlantern/lantern-box/tracker/peerconn"
 	"github.com/getlantern/radiance/common/settings"
 	"github.com/getlantern/radiance/events"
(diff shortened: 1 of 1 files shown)


teamlantern:
This PR has been inactive for 31 days.

Waiting on: none

It will be closed automatically in 7 days if no action is taken.

myleshorton:
Superseded by #591, which delivers this functionality on top of the peer-core integration branch.

Worth recording why this could not simply be rebased: the six-line diff here compiles against current `main`, passes CI, and **silently does nothing**.

`box.BaseContext()` constructs a fresh `service.Registry` on every call, and `boxRegistryCtx.Value` called it on every cache miss — so each lookup resolved to a different registry object. `service.MustRegister` mutates whichever registry the context returns, so the registration landed in an object discarded before libbox read it. Reads were unaffected, because every fresh base carries the same protocol registrations; only a *registration* could expose the problem, and until this PR nothing registered.

#591 captures the base context once per box, registers the factory against it, and resolves values from that single instance while still taking cancellation from the caller — the shape `vpn/tunnel.go` already uses for the main tunnel. It also adds four tests; the two that encode this bug were confirmed to fail against the old per-lookup rebuild.

Two other things carried over: the `slog` routing rationale is preserved (the Freshdesk reference is dropped, per this repo's CLAUDE.md guidance that tickets live in git history), and this PR's base branch `fisk/peer-connection-events` was the orphaned pre-`-A` branch, 97 commits behind `main`, with a failing `build` check — all of which #591 sidesteps by branching from `smc/peer-core`.

The branch is left intact.

PR #491, opened 2026-05-22
config: fix silently dropped triggered config refetch when already in-flight

garmr-ulfr:
## Summary

A `UserChangeEvent` firing while a poll-driven fetch was in flight could race the atomic dedup guard, be silently skipped, and leave the handler on stale config until the next ~10-minute poll tick — so after a login, the user could remain on pre-login config for up to ten minutes.

Replaces the `atomic.Bool` dedup with a mutex-protected `fetching`/`pending` pair. A caller arriving during an in-flight fetch sets `pending` and returns immediately; the in-flight loop re-runs `doFetchConfig` once after the current one returns, picking up the freshest identity for the second request. Worst case after an event storm: one extra HTTP round-trip on the stale-identity fetch, then a fresh one.

### DIFF
--- a/backend/radiance.go
+++ b/backend/radiance.go
@@ -294,7 +294,7 @@ func (r *LocalBackend) Start() {
 			slog.Error("Failed to run offline URL tests after config update", "error", err)
 		}
 	})
-	go r.confHandler.Start()
+	r.confHandler.Start()
 }
 
 func (r *LocalBackend) Close() {
--- a/config/config.go
+++ b/config/config.go
@@ -68,15 +68,24 @@ type Options struct {
 // to the most recent configuration.
 type ConfigHandler struct {
 	// config holds a configResult.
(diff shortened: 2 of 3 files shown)


myleshorton:
Nice! I think this also means that callbacks associated with that config won't happen right? Which will mess up the bandit.

myleshorton (approved):

WendelHime (approved):

garmr-ulfr:
> I think this also means that callbacks associated with that config won't happen right? Which will mess up the bandit.

That's a good question. Are you referring to the offline URLTests?

myleshorton:
I'm just saying that, without this fix, the dropped config refetch will also mean that callbacks associated with that config won't happen. Either way, though, we should pull this in.

garmr-ulfr:
Oh, OK, that makes sense. Pulling it in.

PR #497, opened 2026-05-28
feat: enabling dnstt

WendelHime:
This pull request introduces significant improvements to the DNSTT (DNS tunnel) transport in the `kindling` subsystem, enhances reliability and performance under censorship, and updates several dependencies. The DNSTT transport is now enabled by default, features more robust tunnel management and probing, and better integrates with the system's bypass logic. Additionally, HTTP timeouts have been increased, and dependencies have been updated for improved stability and compatibility.

**DNSTT Transport Improvements:**

- The DNSTT transport is now enabled by default in `EnabledTransports`, making it available for use in production scenarios.
- The DNSTT transport now creates and probes tunnel instances on demand, rather than at startup, reducing resource usage and improving resilience. Tunnel management has been refactored to track consecutive failures, discard failing tunnels, and periodically probe for working tunnels. [[1]](diffhunk://#diff-5354bf3a34515c0a31dd81daa5951a82d9a34881d05123c30625d4006ec7a606L84-R93) [[2]](diffhunk://#diff-5354bf3a34515c0a31dd81daa5951a82d9a34881d05123c30625d4006ec7a606L250-R493)
- The DNSTT transport now uses the `bypass.DialContext` dialer, ensuring that DoH/DoT connections do not loop back through the VPN tunnel.
- Added `RequestTimeout` and `MaxLength` methods to DNSTT transport, tuning request timeouts and maximum body size for the high-latency, low-MTU nature of DNS tunnels.
- Improved tunnel cleanup and cancellation logic, ensuring that probe workers are promptly stopped and successful tunnels are properly closed.

**Kindling Client and Transport Integration:**

- The logic for enabling and configuring transports in `NewKindling` has been updated, moving the Smart transport configuration before DNSTT and ensuring correct option ordering. [[1]](diffhunk://#diff-e3a6647b817184288f0d27bc9e0462c083a84c3381841c0f24623a1611db6f27R153-R158) [[2]](diffhunk://#diff-e3a6647b817184288f0d27bc9e0462c083a84c3381841c0f24623a1611db6f27L165-L170)

**General Configuration and Constants:**

- Increased the default HTTP timeout from 60 seconds to 360 seconds to accommodate slow connections, especially over DNS tunnels.

**Dependency Updates:**

- Updated several dependencies in `go.mod`, including `github.com/getlantern/dnstt`, `go.opentelemetry.io/otel`, `golang.org/x/crypto`, and others, to newer versions for improved performance and compatibility. [[1]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L31-R35) [[2]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L51-R59) [[3]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L217-R230) [[4]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L289-R310)
- Updated test assertions in `parser_test.go` to reflect the refactored DNSTT transport internals (`configs` instead of `options`). [[1]](diffhunk://#diff-08f623607f6676a2aa08f5ffea88cd6ddbf86dea67954db088ce25cfa9dcf9b9L135-R135) [[2]](diffhunk://#diff-08f623607f6676a2aa08f5ffea88cd6ddbf86dea67954db088ce25cfa9dcf9b9L150-R150) [[3]](diffhunk://#diff-08f623607f6676a2aa08f5ffea88cd6ddbf86dea67954db088ce25cfa9dcf9b9L167-R167) [[4]](diffhunk://#diff-08f623607f6676a2aa08f5ffea88cd6ddbf86dea67954db088ce25cfa9dcf9b9L178-R178)

**Code Structure and Typing:**

- Added a type alias for `DNSTT` in the parser to simplify imports and usage.
- Refactored tunnel configuration and resolver logic for clarity and maintainability.

These changes collectively make the DNSTT transport more robust, production-ready, and easier to maintain, while also keeping the codebase up-to-date with dependency improvements.


common/constants.go
```diff
@@ -14,7 +14,7 @@ var Version = "dev"
 const (
 	Name = "lantern"
 
-	DefaultHTTPTimeout = (60 * time.Second)
+	DefaultHTTPTimeout = (360 * time.Second)
```
garmr-ulfr: This is used by more than DNSTT, so we should probably define a separate variable for DNSTT timeout.

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,240 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
-	if len(m.options) < poolSize {
-		poolSize = len(m.options)
+	if len(m.configs) < poolSize {
+		poolSize = len(m.configs)
 	}
 
 	pool := pond.New(poolSize, 10, pond.Context(pondCtx))
-	for _, dnst := range m.options {
+	for _, cfg := range m.configs {
+		cfg := cfg
```
garmr-ulfr: `cfg := cfg` not necessary since go v1.22.

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,240 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
+				tun.Close()
+				return
 			}
 
-			req, err := http.NewRequestWithContext(pondCtx, http.MethodGet, "https://www.gstatic.com/generate_204", http.NoBody)
+			// 180 s covers DNSTT session establishment (~20-60 s over DoH) and
+			// TLS handshake through 135-byte MTU tunnel (multiple round trips).
+			client := &http.Client{Transport: rt, Timeout: 180 * time.Second}
+
+			req, err := http.NewRequestWithContext(pondCtx, http.MethodGet, "http://www.gstatic.com/generate_204", http.NoBody)
```
garmr-ulfr, line 352: Did you mean to change to HTTP instead of HTTPS? Since one issue was that tunnels timeout/fail during the TLS handshake (due to transfer speed, packet size, etc.), we should use a request equivalent to what will be sent when actually running. Otherwise, the test is kind of meaningless.

kindling/dnstt/parser.go
```diff
@@ -81,21 +86,11 @@ func DNSTTOptions(ctx context.Context, localConfigFilepath string, logger io.Wri
...
-			slog.Warn("failed to build dnstt", slog.Any("error", err))
-			continue
-		}
-
-		tunnels = append(tunnels, &dnsTunnel{DNSTT: dnst})
-	}
-
 	m := &multipleDNSTTTransport{
-		tunChan:  make(chan *dnsTunnel, 400),
+		tunChan: make(chan *dnsTunnel, 400),
```
garmr-ulfr: 400 is kind of high, especially given the recent iOS memory issues. Since `tunChan` holds successful tunnels, we only need to retain a few, <10. Once we find enough working ones we can stop testing the rest, then only we start testing again from were we left off to find replacements for ones that stop working upto the max.

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,240 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
-			slog.Debug("adding successful tun to channel")
-			dnst.markSucceeded()
-			if !m.closed.Load() {
-				m.tunChan <- dnst
-			}
+		slog.Debug("dnstt tunnel ready", slog.String("domain", cfg.Domain), slog.String("resolver", resolver))
+		tun.markSucceeded()
+		if !m.closed.Load() {
+			m.tunChan <- tun
+		} else {
```
garmr-ulfr: This builds and adds a new DNSTT instance for successful ones but they're only removed when they fail. So eventually `tunChan` will fill up and no more can be added, resulting in this blocking until one does fail and is removed and leaking if that doesn't happen. We should skip finding new tunnels if we have some working ones already (5 is reasonable) and only replace broken ones. If we want to establish new tunnels when we receive a new config, we should completely replace existing ones, while not exceeding the max open tunnels.

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,240 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
 	defer t.mx.RUnlock()
-	return t.lastSucceeded.After(time.Time{})
+	return t.lastSucceeded.After(time.Time{}) && t.consecutiveFailures < maxTunnelFailures
 }
 
-// NewRoundTripper creates a new HTTP round tripper for the given address.
-// It manages session creation and reuse.
+// NewRoundTripper creates a pre-connected HTTP round tripper for the given
+// address. It blocks until a KCP session and smux stream are established so
+// that the race transport can fairly compare connection latencies.
```
garmr-ulfr: > It blocks until a KCP session and smux stream are established so that the race transport can fairly compare connection latencies.

The KCP session and smux streams are actually established lazily — on first request. So this isn't correct.

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,240 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
 			if !tun.isSucceeding() {
+				tun.Close()
+				continue
+			}
+			rt, err := tun.NewRoundTripper(ctx, addr)
+			if err != nil {
+				tun.recordFailure()
+				tun.Close()
+				slog.WarnContext(ctx, "dnstt roundtripper creation failed during connect",
+					"domain", tun.domain, "resolver", tun.resolver, "error", err)
```
garmr-ulfr: `NewRoundTripper` only returns an error if the dnstt instance is already closed. So there's no need to double-close it here or record the failure.

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,240 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
-	if len(m.options) == 0 {
+	if len(m.configs) == 0 {
 		slog.Debug("no dns tunnel options available")
 		return
 	}
 
 	pondCtx, cancel := context.WithTimeout(context.Background(), waitFor)
+	m.probeCancelMx.Lock()
+	m.probeCancelFn = cancel
+	m.probeCancelMx.Unlock()
```
garmr-ulfr, line 309: `tryAllDNSTunnels` is triggered by `findWorkingDNSTunnels` (immediately), `NewRoundTripper` (via `probeCh`), and by the timer, which can overlap. Since `probeCancelFn` is overwritten here each time, it can only cancel the most recently triggered probe. We should use a flag to signal that it's already running and prevent subsequent runs from being triggered.

garmr-ulfr (changes requested):

kindling/dnstt/parser.go
WendelHime: Yes, I've changed to HTTP because there was too many failures with dns tunnels while using HTTPS for testing and this test basically create a session and keep it open to further requests to just use the same open session

garmr-ulfr: Oh OK. Is that already documented somewhere? If not, would you add a comment in the function doc explaining that this is used to eagerly establish connections?

kindling/dnstt/parser.go
```diff
@@ -247,183 +257,276 @@ func parseDNSTTConfigs(gzipyml []byte) ([]dnsttConfig, error) {
...
-
-			// Add the tun back to the channel.
-			m.tunChan <- tun
-			return &connectedRoundtripper{t: tun, ctx: ctx, addr: addr}, nil
+			// Any error here means the tunnel is unusable, so drop it and try
+			// the next rather than closing it again or recording a failure.
+			rt, err := tun.NewRoundTripper(ctx, addr)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
WendelHime: tun.NewRoundTripper only returns an error when the underlying dnstt instance is already closed

garmr-ulfr (approved):
LGTM!

PR #509, opened 2026-06-04
set default TUN stack to system

garmr-ulfr:
Use the system stack for the TUN for all platforms except Android devices with kernels <=v5.10. This will avoid using gvisor for UDP connections, which will reduce memory usage. And as a side-bonus, improve performance.


garmr-ulfr:
Pulling this in.

PR #512, opened 2026-06-07
unbounded: keep consumer Source on close events (fix orphaned globe arcs + runaway helped counter)

myleshorton:
## Problem

The Unbounded donor tab's **"People being helped right now"** counter runs away (a tester saw **131** while the pill still said "Waiting for connections…"), and it equals "Total people helped to date" — the live count only ever grows. The globe also accumulates arcs that never disappear.

Root cause is here in `unbounded.go`. broflake's `OnConnectionChangeFunc` hands us the consumer's `addr` on **accept** but a **nil `addr` on close** (the WebRTC session is already torn down, so the remote IP is gone). We emit `ConnectionEvent{Source: addr.String()}`, so a close goes out with `Source == ""`.

Downstream, both consumers of the event identify a connection **by its Source IP**:
- the Flutter globe matches a close to the arc it should remove by source IP;
- the "people helped" counter decrements by source IP.

So an empty-Source close can't be matched — the arc orphans and the counter never comes back down. broflake *does* fire the close (it has to, to recycle its small consumer table), so these aren't missing events; they just arrive un-attributable. (The 131 magnitude was amplified by a separate consumer-routing flood, throttled on the lantern-cloud side.)

## Fix

Track each consumer slot's addr on accept and restore it on close, so every `-1` carries the same `Source` its `+1` did:

- `connSources` — a small `workerIdx → addr` map (broflake's `workerIdx` is the consumer slot, stable across a single connection's accept→close), concurrency-safe since broflake fires callbacks from per-worker goroutines.
- `OnConnectionChangeFunc` calls `sources.resolve(state, workerIdx, addrStr)` — records on accept, restores on close.

No change to the `ConnectionEvent` shape or to any consumer; they already key off `Source`. An accept that genuinely has no addr stays empty through its close (neither end is counted), which is the correct behavior when broflake can't surface the consumer IP at all.

## Tests
- `TestConnSources_resolve` — unit-tests the backfill: accept stores, nil-addr close restores, freed slot restores nothing, slots independent, empty-addr accept stays empty, real-addr close passes through, slot reuse.
- `TestConnectionEventBridge` — updated to assert the installed callback now emits the accept's IP on the close (was asserting the old empty-Source bug). Passes with `-race`.

### DIFF
--- a/unbounded/conn_sources_test.go
+++ b/unbounded/conn_sources_test.go
@@ -1,0 +1,56 @@
+package unbounded
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestConnSources_resolve pins the accept→close source backfill. broflake
+// delivers the consumer addr on accept but a nil (empty) addr on close, so
+// resolve must restore the accept's addr onto the close — otherwise the close
+// event carries an empty Source and downstream consumers (the Flutter globe
+// arc + the "people helped" counter) can't match it to the accept, so the
+// arc orphans and the counter never decrements.
+func TestConnSources_resolve(t *testing.T) {
+	c := newConnSources()
+
+	// Accept echoes its own addr and remembers it for the slot.
+	assert.Equal(t, "1.2.3.4", c.resolve(1, 7, "1.2.3.4"),
+		"accept returns its own source")
+
+	// Close arrives with an empty addr (broflake's nil) — restore the
+	// accept's addr so the -1 can be matched to its +1.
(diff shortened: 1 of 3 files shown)


myleshorton:
Superseded by #590, which carries these commits verbatim.

Keeping the consumer `Source` on close events — the `connSources` worker-index → address resolution that fixes orphaned globe arcs and the runaway helped counter — is in #590 unchanged, and `go test ./unbounded/...` passes there.

The branch is left intact — nothing is deleted, and this can be reopened if any of the above is wrong.

PR #527, opened 2026-06-15
Add novpn build variants for tunnel and split tunnel

reflog:
Introduce build-tagged vpn implementations that remove TUN and
split-tunnel
behavior in `novpn` builds while preserving API compatibility. Add
`novpn` stubs for split tunnel, force SOCKS-only mode with required
address validation, and split inbound/split-tunnel option construction
into default vs novpn files. Move shared split-tunnel types/constants to
a common file and gate affected tests with build tags, adding novpn-
specific coverage.


vpn/socks_only_novpn.go
```diff
@@ -0,0 +1,17 @@
...
+// device. A missing RADIANCE_SOCKS_ADDRESS is unrecoverable misconfiguration for
+// this build, so it panics rather than starting a tunnel that can never carry
+// traffic.
+func socksOnlyEnforced() bool {
+	if addr, ok := env.Get(env.SocksAddress); !ok || addr == "" {
+		panic("novpn build requires " + env.SocksAddress.String() + " to be set")
+	}
+	env.Set(env.UseSocks.String(), "true")
+	return true
+}
```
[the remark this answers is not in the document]
reflog: Agreed. Updated `socksOnlyEnforced` to use `os.Setenv` so enforced SOCKS mode overrides any existing process environment value, and it now panics if that process env write fails.

vpn/socks_only_novpn_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/getlantern/radiance/common/env"
+)
+
+// TestMain seeds a SOCKS address so the many buildOptions-based tests in this
+// package don't trip socksOnlyEnforced's mandatory-address panic.
+func TestMain(m *testing.M) {
+	env.Set(env.SocksAddress.String(), "127.0.0.1:1080")
```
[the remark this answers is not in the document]
reflog: Agreed. `TestMain` now seeds `RADIANCE_SOCKS_ADDRESS` via `os.Setenv`, and the force test explicitly starts with `RADIANCE_USE_SOCKS_PROXY=false` to cover OS env precedence.

vpn/boxoptions.go
```diff
@@ -489,7 +430,7 @@ func buildOptions(bOptions BoxOptions) (O.Options, error) {
 	opts := baseOpts(bOptions.BasePath)
 	slog.Debug("Base options initialized")
 
-	if env.GetBool(env.UseSocks) {
+	if socksOnlyEnforced() || env.GetBool(env.UseSocks) {
 		socksAddr, _ := env.Get(env.SocksAddress)
 		slog.Info("Using SOCKS proxy for inbound as per environment variable", "socksAddr", socksAddr)
```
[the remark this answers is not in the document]
reflog: Agreed. Changed the log line to `Using SOCKS proxy for inbound` so it applies to both enforced novpn and opt-in env modes.

vpn/split_tunnel_novpn.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (s *SplitTunnel) IsEnabled() bool { return false }
+
+func (s *SplitTunnel) SetEnabled(bool) error { return nil }
+
+func (s *SplitTunnel) Filters() SplitTunnelFilter { return SplitTunnelFilter{} }
+
+func (s *SplitTunnel) AddItems(SplitTunnelFilter) error { return nil }
+
+func (s *SplitTunnel) RemoveItems(SplitTunnelFilter) error { return nil }
```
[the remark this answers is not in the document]
reflog: Agreed. Changed the intentionally unused stub parameters to blank identifiers while keeping the exported signatures compatible.

vpn/splittunnel_options_novpn.go
```diff
@@ -0,0 +1,9 @@
+//go:build novpn
+
+package vpn
+
+import O "github.com/sagernet/sing-box/option"
+
+func splitTunnelRuleSet(string) []O.RuleSet { return nil }
```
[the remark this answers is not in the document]
reflog: Agreed. Changed the unused parameter to `_ string` so the no-op novpn signature is clear.

vpn/inbounds_novpn.go
```diff
@@ -0,0 +1,34 @@
...
+	return []O.Inbound{
+		{
+			Type: C.TypeMixed,
+			Tag:  bypass.BypassInboundTag,
+			Options: &O.HTTPMixedInboundOptions{
+				ListenOptions: O.ListenOptions{
+					Listen:     &loopbackAddr,
+					ListenPort: bypass.ProxyPort,
+				},
+			},
```
WendelHime: In this case, do we need the [`SetSystemProxy`](https://sing-box.sagernet.org/configuration/inbound/mixed/#set_system_proxy) flag? (for windows, android, linux and mac os)

reflog: @WendelHime it doesn't really work on Android (without root).
And novpn changes are Android only.
So I don't think it's needed?

WendelHime (approved):

vpn/socks_only_novpn.go
```diff
@@ -0,0 +1,23 @@
...
+// this build, so it panics rather than starting a tunnel that can never carry
+// traffic.
+func socksOnlyEnforced() bool {
+	if addr, ok := env.Get(env.SocksAddress); !ok || addr == "" {
+		panic("novpn build requires " + env.SocksAddress.String() + " to be set")
+	}
+	if err := os.Setenv(env.UseSocks.String(), "true"); err != nil {
+		panic("novpn build failed to set " + env.UseSocks.String() + ": " + err.Error())
+	}
+	return true
```
garmr-ulfr: Instead of requiring an environment variable, we could accept the address as a config option with a default, while still allowing the environment variable override it. `UseSocks` also becomes moot for `novpn` as it's the only option.

garmr-ulfr:
Still reviewing but had a couple of questions.
~- Will the OS be configured to route all HTTP traffic through the SOCKS proxy?~
~- Does `novpn` need to run without looking like a VPN? In other words, not using the `VpnService` and therefor not needing the VPN permissions?~

Nevermind. Read the [epic](https://github.com/getlantern/engineering/issues/3569) tied to the corresponding [lantern PR](https://github.com/getlantern/lantern/pull/8861). Would you update the description to include "why" this is change is needed? It's not obvious and required some digging to fully understand.

vpn/inbounds_novpn.go
garmr-ulfr: We could actually remove the bypass inbound — and put the whole bypass pkg behind the `!novpn` build tag, for that matter. It's only purpose is for kindling to bypass the TUN, which won't exist in this case.

vpn/boxoptions.go
```diff
@@ -509,25 +450,7 @@ func buildOptions(bOptions BoxOptions) (O.Options, error) {
...
-			slog.Debug("Android platform detected, OverrideAndroidVPN set to true")
-		case "ios":
-			opts.Inbounds[0].Options.(*O.TunInboundOptions).Stack = ""
-			slog.Debug("iOS platform detected, using default TUN stack with no override")
-		case "linux":
-			opts.Inbounds[0].Options.(*O.TunInboundOptions).AutoRedirect = true
-			slog.Debug("Linux platform detected, AutoRedirect set to true")
-		}
+		applyPlatformTunnelOptions(&opts)
 	}
```
garmr-ulfr: I don't think this is this even necessary anymore. We could just move the SOCKS inbound to `inbounds_novpn.baseInbounds`, matching the `inbounds_default` one. This lets us drop both `socks_only_default.go` and `socks_only_novpn.go`, and also the `UseSocks` env var.

vpn/inbounds_default.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		slog.Debug("Android platform detected, OverrideAndroidVPN set to true")
+	case "ios":
+		opts.Inbounds[0].Options.(*O.TunInboundOptions).Stack = ""
+		slog.Debug("iOS platform detected, using default TUN stack with no override")
+	case "linux":
+		opts.Inbounds[0].Options.(*O.TunInboundOptions).AutoRedirect = true
+		slog.Debug("Linux platform detected, AutoRedirect set to true")
+	}
+}
```
garmr-ulfr: We could just set these in `baseInbounds` instead since it's only modifying the TUN inbound options. Then we won't need to define the same function in `inbounds_novpn.go`. We could also move `kernelBelow` and `minAndroidSystemStackKernel` to this file.

vpn/split_tunnel.go
```diff
@@ -1,3 +1,5 @@
+//go:build !novpn
+
```
garmr-ulfr, line 2: Do we need to omit split tunneling for `novpn` entirely? Domain filtering will still work.

vpn/splittunnel_options_default.go
garmr-ulfr, line 1: Since `inbounds_xxx.go` and `splittunnel_options_xxx.go` are just the separated box options code, I think it would more sense to combine them into `boxoptions_xxx.go` files.

garmr-ulfr (commented):
Looks good! Just some clean up/simplification that would be much appreciated.

vpn/socks_only_novpn.go
reflog: Done. The novpn `baseInbounds` now builds the SOCKS inbound directly with a default address (`defaultSocksAddress = 127.0.0.1:1080`) that `RADIANCE_SOCKS_ADDRESS` overrides, and `UseSocks` is gone. I implemented the "config option with a default" as a package default constant + env override rather than a `BoxOptions` field, since nothing plumbs such a field today — happy to promote it to a real config field if you prefer.

vpn/inbounds_novpn.go
reflog: Dropped the bypass inbound from novpn, along with its routing rule (now behind a build-tagged `bypassRoutingRules`). I couldn't put the whole `bypass` package behind `!novpn` though: `kindling` (client/smart/fronted/dnstt) and `servers/manager` use `bypass.StreamDialer`/`Dial`/`DialContext`, which must compile in the novpn build. So only the inbound and its rule are removed, not the package.

vpn/boxoptions.go
reflog: Done. Removed the socks `if/else` from `buildOptions`, deleted `socks_only_{default,novpn}.go` and the `UseSocks` env var; the SOCKS inbound now lives in novpn `baseInbounds`, matching the TUN inbound in `inbounds_default`.

vpn/inbounds_default.go
reflog: Done. The per-platform TUN tweaks are applied inline in the default `baseInbounds`, and `kernelBelow` + `minAndroidSystemStackKernel` moved with it. `applyPlatformTunnelOptions` and its novpn stub are gone. The one non-inbound thing it did — `OverrideAndroidVPN`, a `Route` option — is now set in `baseOpts` gated on `hasTunInbound`, so it stays correct for both builds without a per-build hook.

vpn/split_tunnel.go
reflog: Good point — domain filtering would work in SOCKS mode. For now we opted to keep the inert stub (conservative) and just merge the per-build option files per your other comment. Easy to revisit and wire up the real split-tunnel for novpn later if we want it.

vpn/splittunnel_options_default.go
reflog: Done. `inbounds_{default,novpn}.go` and `splittunnel_options_{default,novpn}.go` are now combined into `boxoptions_{default,novpn}.go`.

vpn/boxoptions_test.go
```diff
@@ -480,74 +480,6 @@ func TestHasGlobalIPv6Using(t *testing.T) {
 // reject: it must follow the split-tunnel and smart-routing rules (so a
 // direct-routed domain keeps its QUIC) and precede the selector rules (so
 // QUIC bound for the proxy is rejected).
```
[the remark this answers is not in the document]
reflog: Done. Renamed the doc comment to TestBuildOptions_RejectsIPv6WhenCaptured and rewrote it to describe the ::/0 reject ordering (after split-tunnel, before selectors, only when the TUN captured a v6 address). The stale QUIC text is gone.

vpn/boxoptions.go
```diff
@@ -323,49 +284,23 @@ func baseRoutingRules() []O.Rule {
...
-					IPIsPrivate: true,
-				},
-				RuleAction: O.RuleAction{
-					Action: C.RuleActionTypeRoute,
-					RouteOptions: O.RouteActionOptions{
-						Outbound: "direct",
-					},
-				},
+	}
+	rules = append(rules, bypassRoutingRules()...)
```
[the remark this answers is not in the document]
reflog: Done. Added an inline note at the call site that bypassRoutingRules() returns nil in novpn builds (no TUN, so kindling's direct-routing carve-out is unnecessary).

vpn/boxoptions.go
```diff
@@ -323,49 +284,23 @@ func baseRoutingRules() []O.Rule {
...
+			RuleAction: O.RuleAction{
+				Action: C.RuleActionTypeRoute,
+				RouteOptions: O.RouteActionOptions{
+					Outbound: "direct",
 				},
 			},
 		},
-	}
+	})
+	rules = append(rules, splitTunnelRoutingRules()...)
```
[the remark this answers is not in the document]
reflog: Done. Added an inline note at the call site that splitTunnelRoutingRules() returns nil in novpn builds, which have no split tunnel.

vpn/boxoptions_novpn.go
```diff
@@ -0,0 +1,52 @@
...
+	if v, ok := env.Get(env.SocksAddress); ok && v != "" {
+		addr = v
+	}
+	addrPort, err := netip.ParseAddrPort(addr)
+	if err != nil {
+		slog.Warn("invalid SOCKS address, using default",
+			"address", addr, "default", defaultSocksAddress, "error", err)
+		addrPort = netip.MustParseAddrPort(defaultSocksAddress)
+	}
+	listen := badoption.Addr(addrPort.Addr())
```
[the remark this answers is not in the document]
reflog: Agreed. baseInbounds now rejects non-loopback or zero-port addresses and falls back to the default (127.0.0.1:1080), so a misconfigured RADIANCE_SOCKS_ADDRESS can't expose the proxy to the LAN.

common/env/env.go
```diff
@@ -24,7 +24,6 @@ var (
 	PrintCurl        _key = "RADIANCE_PRINT_CURL"
 	DisableStdout    _key = "RADIANCE_DISABLE_STDOUT_LOG"
 	ENV              _key = "RADIANCE_ENV"
-	UseSocks         _key = "RADIANCE_USE_SOCKS_PROXY"
 	SocksAddress     _key = "RADIANCE_SOCKS_ADDRESS"
 	Country          _key = "RADIANCE_COUNTRY"
```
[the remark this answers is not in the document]
reflog: Done. Removed the remaining RADIANCE_USE_SOCKS_PROXY references: the README row is gone (replaced by an updated RADIANCE_SOCKS_ADDRESS entry describing the novpn SOCKS inbound), and tester/Dockerfile now builds with the novpn tag instead of setting the env var. So the documented surface matches the code: novpn always uses SOCKS, configured via RADIANCE_SOCKS_ADDRESS.

garmr-ulfr (approved):
LGTM! Thanks for the cleanup!

PR #530, opened 2026-06-16
Added an event for a URL test.

jigar-f:
This pull request introduces a new event and event stream for notifying clients when URL test runs (either offline or online) complete and produce usable latency results. The main changes revolve around defining the new event type, emitting these events from the backend, and exposing them via both in-process and SSE APIs.

**Key changes include:**

### Event Definition and Emission

* Added a new `URLTestCompleteEvent` type to `vpn/vpn.go`, including constants to distinguish between offline and online test sources.
* Modified `backend/radiance.go` to emit `URLTestCompleteEvent` after both online and offline URL test runs when results are available. [[1]](diffhunk://#diff-d7a5115ab346da59601a056266b297ad60b2db8ce5434ba7cd67023d47d5097bR724) [[2]](diffhunk://#diff-d7a5115ab346da59601a056266b297ad60b2db8ce5434ba7cd67023d47d5097bR978)

### API and Event Streaming

* Introduced a new SSE endpoint `/server/url-test/events` in `ipc/server.go` and implemented the corresponding handler to stream `URLTestCompleteEvent` notifications to clients. [[1]](diffhunk://#diff-dc406fc743858a2874d0cebfbc7c0be932056c3a247189aca780ca5ac3342c64R49) [[2]](diffhunk://#diff-dc406fc743858a2874d0cebfbc7c0be932056c3a247189aca780ca5ac3342c64R215) [[3]](diffhunk://#diff-dc406fc743858a2874d0cebfbc7c0be932056c3a247189aca780ca5ac3342c64R533-R560)
* Added a `URLTestEvents` method to the client API in both `ipc/client_events_mobile.go` and `ipc/client_events_nonmobile.go`, allowing clients to subscribe to URL test completion events via either the in-process bus or SSE, depending on context. [[1]](diffhunk://#diff-cf9948a518ce45d66edeb9a07eafb29590f293af70b57e27bd3aaff9d6609218R32-R50) [[2]](diffhunk://#diff-ac126e34c8a04a78b7c42c6645cfbdff0fe93042f117e5106582c99221f8bea7R23-R34)

### DIFF
--- a/backend/radiance.go
+++ b/backend/radiance.go
@@ -721,6 +721,7 @@ func (r *LocalBackend) flushSelectionHistory(storage vpn.AutoSelectHistoryStorag
 		if err := r.srvManager.UpdateSelectionHistory(results); err != nil {
 			slog.Warn("Failed to persist selection history", "error", err)
 		}
+		events.Emit(vpn.URLTestCompleteEvent{Source: vpn.URLTestSourceOnline, Count: len(results)})
 	}
 }
 
@@ -974,6 +975,7 @@ func (r *LocalBackend) RunOfflineURLTests() error {
 		if err := r.srvManager.UpdateSelectionHistory(histories); err != nil {
 			slog.Warn("Failed to persist offline selection history", "error", err)
 		}
+		events.Emit(vpn.URLTestCompleteEvent{Source: vpn.URLTestSourceOffline, Count: len(histories)})
 		selected, err := r.vpnClient.CurrentAutoSelectedServer()
 		if err != nil {
 			slog.Warn("Failed to get current auto-selected server after URL tests", "error", err)
--- a/backend/radiance.go
+++ b/backend/radiance.go
@@ -691,29 +691,26 @@ func (r *LocalBackend) updateSelectionHistoryListener(status vpn.VPNStatus) {
 func (r *LocalBackend) runSelectionHistoryListener(ctx context.Context, storage vpn.AutoSelectHistoryStorage, hook <-chan struct{}) {
 	ticker := time.NewTicker(selectionHistoryFlushInterval)
 	defer ticker.Stop()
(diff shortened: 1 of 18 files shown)


backend/radiance.go
```diff
@@ -691,26 +691,29 @@ func (r *LocalBackend) updateSelectionHistoryListener(status vpn.VPNStatus) {
 func (r *LocalBackend) runSelectionHistoryListener(ctx context.Context, storage vpn.AutoSelectHistoryStorage, hook <-chan struct{}) {
 	ticker := time.NewTicker(selectionHistoryFlushInterval)
 	defer ticker.Stop()
-	dirty := true // start dirty so we persist any results that arrived before the listener started
+	dirty := true   // start dirty so a seed/result present before the listener started still gets persisted
+	probed := false // that seed flush is not a probe outcome; emit only after a real probe write
```
garmr-ulfr: The online test hook fires on any history change — not only when a probe was run but also when any of the other 4 tracked metrics change — so this could fire dozens of times between actual probe tests, impacting performance. There isn't a way to listen for online probe tests specifically.

backend/radiance.go
```diff
@@ -974,6 +974,7 @@ func (r *LocalBackend) RunOfflineURLTests() error {
 		if err := r.srvManager.UpdateSelectionHistory(histories); err != nil {
 			slog.Warn("Failed to persist offline selection history", "error", err)
 		}
+		events.Emit(vpn.URLTestCompleteEvent{Source: vpn.URLTestSourceOffline, Count: len(histories)})
```
garmr-ulfr: If you think it would be useful for the app-side, we could also add the results in the event.

garmr-ulfr (approved):
LGTM!

backend/radiance.go
jigar-f: Thanks, I added, result.

PR #531, opened 2026-06-16
extend offline URL test pre-warm timeout

atavism:
## Summary
- Increase the offline URL-test pre-warm timeout from 5s to 30s.
- Log the outbound count and timeout when offline URL tests start.

## Why
Radiance runs offline URL tests through a temporary `mutableautoselect` group, which probes outbounds with bounded concurrency. With the previous 5s total timeout, larger server pools could be cancelled before every queued probe had a chance to run, making servers appear unavailable even though they had not been fully tested.


garmr-ulfr (approved):
LGTM!

PR #538, opened 2026-06-24
Fix retained servers missing from cold-start VPN options

atavism:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved managed server option assembly to consistently retain existing outbound/endpoint entries while avoiding duplicates via tag-based deduplication.
  * Ensured retained and user server options remain connectable for cold-start by merging managed options without overwriting existing “current” settings (including outbound type).
* **Tests**
  * Added coverage for merging managed lantern and non-lantern server options, including endpoint alias retention and conflict behavior preservation.

### DIFF
--- a/backend/radiance.go
+++ b/backend/radiance.go
@@ -857,16 +857,11 @@ func (r *LocalBackend) getBoxOptions() vpn.BoxOptions {
 			bOptions.AdBlock = cfg.AdBlock
 		}
 	}
+	servers := r.srvManager.AllServers()
+	appendManagedServerOptions(&bOptions.Options, servers)
+
 	seed := make(map[string]lbA.TagHistory)
-	for _, srv := range r.srvManager.AllServers() {
-		if !srv.IsLantern {
-			switch opts := srv.Options.(type) {
-			case option.Outbound:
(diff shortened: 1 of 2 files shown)


backend/radiance.go
```diff
@@ -877,6 +872,63 @@ func (r *LocalBackend) getBoxOptions() vpn.BoxOptions {
...
+		switch opts := srv.Options.(type) {
+		case option.Outbound:
+			tag := managedOptionTag(srv.Tag, opts.Tag)
+			if tag == "" {
+				continue
+			}
+			if _, exists := existingTags[tag]; exists {
+				continue
+			}
+			opts.Tag = tag
```
garmr-ulfr: `srv.Tag` and `opt.Tag` are guaranteed to be equal and non-empty. Checking if not empty harmless but `opt.Tag` is the authoritative tag and re-assigning can hide bugs. We should use the values already set.

garmr-ulfr (changes requested):

backend/radiance.go
atavism: Makes sense! Updated to use the option tag directly and removed the fallback/reassignment

garmr-ulfr (approved):
LGTM!

PR #543, opened 2026-06-29
Apply cached config during startup

atavism:
Fixes a startup race where cached config could be loaded before config-event subscribers were attached, leaving cached servers unavailable if a fresh config fetch failed.

Resolves https://github.com/getlantern/engineering/issues/3640

## Summary by CodeRabbit

* **New Features**
  * Radiance now restores cached on-disk configuration during startup, bringing saved server settings online earlier.
* **Bug Fixes**
  * Configuration refreshes now apply updates via a centralized flow that better updates the managed server list.
  * Improved country handling to correctly persist/propagate saved country data only when appropriate, enhancing server location restoration.
* **Tests**
  * Added an integration-style test validating cached configuration loading and verifying expected server/tag and location fields.


backend/radiance.go
```diff
@@ -225,65 +225,101 @@ func (r *LocalBackend) Start() {
...
-						attribute.Int("bandit.outbound_count", len(cfg.Options.Outbounds)),
-					),
-				)
-				span.End() // point-in-time marker — config was received at this timestamp
-			}
+		r.applyConfig(evt.New)
+		r.prewarmOfflineURLTests("config update")
+	})
+	if r.applyCurrentConfig() {
+		r.prewarmOfflineURLTests("cached config")
```
garmr-ulfr: The offline tests are blocking so we should run this in a separate goroutine.

backend/radiance.go
```diff
@@ -667,15 +703,22 @@ const maxRetainedLanternServers = 60
...
+		prev, exists := existingByTag[srv.Tag]
+		if !exists {
+			incomingNewCount++
+			return false
+		}
+		if !prev.IsLantern {
+			return true
+		}
+		preserveConfigServerState(srv, prev)
+		return false
```
garmr-ulfr: We should only update a server's config when the tag changes. Since the tag uniquely identifies a config, any change results in a new tag. Some config changes also affect how the protocol processes packets, invalidating the existing selection history. Let's keep the existing `updateServers` behavior.

garmr-ulfr (changes requested):

garmr-ulfr (approved):
LGTM!

PR #545, opened 2026-06-30
Fix IPv6 fake-IP routing for dual-stack DNS

atavism:
Proposed fix for https://github.com/getlantern/engineering/issues/3649

Routes both A and AAAA queries through dual-stack fake-IP so domain rules can handle IPv6 DNS results before the IPv6 reject fallback.

Server DNS overrides also keep this fake-IP fallback

## Summary by CodeRabbit

* **New Features**
  * Added dual-stack fake-IP DNS handling so both IPv4 (A) and IPv6 (AAAA) queries can be routed consistently.
  * DNS fallback is applied without changing the existing DNS rules structure.
* **Bug Fixes**
  * Improved DNS merge behavior to avoid unintended mutation and to apply the correct fallback when overwriting DNS settings.
  * Refined IPv6 rejection ordering to block proxied IPv6 more reliably.
* **Documentation**
  * Updated routing-order guidance to clarify rule evaluation placement.
* **Tests**
  * Expanded coverage for dual-stack fake-IP DNS behavior and updated IPv6 ordering assertions.

### DIFF
--- a/vpn/boxoptions.go
+++ b/vpn/boxoptions.go
@@ -245,8 +245,7 @@ func baseRoutingRules() []O.Rule {
 	//       here so split-tunnel, smart-routed, and config-routed direct paths keep
 	//       their QUIC. Added in buildOptions.
 	// 8.    Reject IPv6 (::/0), only when the TUN captured a v6 address; placed here so
-	//       direct-routed v6 is preserved while non-fake-IP proxied v6 fails fast. Added
-	//       in buildOptions.
+	//       direct-routed v6 is preserved while remaining IPv6 fails fast. Added in buildOptions.
 	// 9,10. Group rules for auto and manual selector modes (added in buildOptions).
 	// 11.   Catch-all blocking rule (added in buildOptions). Traffic not covered
 	//       by previous rules must not automatically bypass the VPN.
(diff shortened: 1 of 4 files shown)


garmr-ulfr:
> Routes both A and AAAA queries through dual-stack fake-IP so domain rules can handle IPv6 DNS results before the IPv6 reject fallback.

We need to reject IPv6 DNS queries as well. Since the fakeip server doesn't actually do any lookup, sing-box sets the request destination to the fully-qualified domain, so it won't be caught by the IPv6 reject routing rule. 

> Server DNS overrides also keep this fake-IP fallback

The sing-box options in the config should be treated as the authoritative source. DNS and routing options should be applied as specified, unless the options depend on something the server cannot know; such as whether the client has IPv6 or not. If the fakeip server and rule aren't included, that would be intentional or a bug in lantern-cloud.

atavism:
@garmr-ulfr Sorry, the original auto-diagnosis looked plausible from the code: there’s a broad `::/0` reject rule, base fake-IP was IPv4-only, and the report was about X failing on a dual-stack macOS network after the IPv6 reject change. Thanks for the log review. Closing

garmr-ulfr:
@atavism no worries — honestly the auto-diagnosis was reasonable given the code; configuring fake-IP for IPv6 really does seem like it should fix this. I came to the same conclusion when the issue first came up, so Adam added it to the config server-side. When that didn't work, I had to dive into the deep rabbit hole that is sing-box to figure out what was going on. And that was pre-Claude so I had to trace it manually; wouldn't recommend it, made my brain hurt :laughing:. Sing-box has so many gotchas that aren't documented anywhere, it's ridiculous. It was a full day before I remembered that our proxies don't even support IPv6 and we could just block it entirely.

PR #558, opened 2026-07-08
fix: skipping amp transport based on country

WendelHime:
## Summary by CodeRabbit

* **New Features**
  * Transport settings can now be enabled/disabled via a public transport toggle API, with changes taking effect after the next rebuild.
  * Country-based transport policy is now applied immediately at startup and continues updating live while running (AMP enabled for all countries except China).

* **Bug Fixes**
  * Cached configuration loading now correctly reapplies the transport policy to match the resolved country and environment overrides.

* **Tests**
  * Added coverage for transport enable/disable behavior.

### DIFF
--- a/backend/radiance.go
+++ b/backend/radiance.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"slices"
+	"strings"
 	"sync"
 
 	"time"
@@ -223,11 +224,12 @@ func (r *LocalBackend) Start() {
 	r.startAutoSelectedListener()
 	r.startSessionAutoSelectListener()
 
-	// Track the country on every config response, not just the first: it feeds
-	// issue reports and gates which transports kindling wires up (AMP is
-	// disabled in China), and the country can change while the app runs.
-	events.SubscribeContext(r.ctx, func(evt config.NewConfigEvent) {
+	// The server derives the country from the client IP, so it's stable for the
+	// session: react once to record it for issue reports and to apply the
+	// country-specific transport policy (AMP is disabled in China).
+	events.SubscribeOnce(func(evt config.NewConfigEvent) {
 		setCountryCodeFromConfig(evt.New)
+		applyTransportPolicy()
 	})
 	// update VPN outbounds when new config is received
 	events.SubscribeContext(r.ctx, func(evt config.NewConfigEvent) {
@@ -248,6 +250,7 @@ func (r *LocalBackend) applyCurrentConfig() bool {
 		return false
 	}
 	setCountryCodeFromConfig(cfg)
+	applyTransportPolicy()
 	r.applyConfig(cfg)
 	return true
 }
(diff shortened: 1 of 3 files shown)


kindling/client.go
```diff
@@ -182,7 +189,7 @@ func NewKindling(dataDir string) (*Client, error) {
 		}
 	}
 
-	if enabled := EnabledTransports[kindling.TransportAMP]; enabled {
+	if EnabledTransports[kindling.TransportAMP] && AMPEnabledForCountry(settings.GetString(settings.CountryCodeKey)) {
```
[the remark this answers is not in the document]
WendelHime: fixed at [513f973](https://github.com/getlantern/radiance/pull/558/commits/513f973358bed11eb3a0615fc3d3255027649b87)

backend/radiance.go
```diff
@@ -223,8 +223,10 @@ func (r *LocalBackend) Start() {
 	r.startAutoSelectedListener()
 	r.startSessionAutoSelectListener()
 
-	// set country code in settings when new config is received so it can be included in issue reports
-	events.SubscribeOnce(func(evt config.NewConfigEvent) {
+	// Track the country on every config response, not just the first: it feeds
+	// issue reports and gates which transports kindling wires up (AMP is
+	// disabled in China), and the country can change while the app runs.
+	events.SubscribeContext(r.ctx, func(evt config.NewConfigEvent) {
```
garmr-ulfr: The country should always be the same on every response — the server uses the client's actual IP, which shouldn't change. All subsequent fired events will spawn goroutines unnecessarily.

backend/radiance.go
```diff
@@ -288,16 +290,36 @@ func (r *LocalBackend) applyConfig(cfg *config.Config) {
...
+	}
 	if err := settings.Set(settings.CountryCodeKey, cfg.Country); err != nil {
 		slog.Error("failed to set country code in settings", "error", err)
+		return
 	}
 	slog.Info("Set country code from config", "country_code", cfg.Country)
+	if kindling.AMPEnabledForCountry(prev) != kindling.AMPEnabledForCountry(cfg.Country) {
+		kindling.Close()
+		kindling.Init()
+	}
```
garmr-ulfr: I think we should move this to the event callback, after the call to `setCountryCodeFromConfig`. `setCountryCodeFromConfig` is a helper function meant to only update the settings, so it seems out of place here.

kindling/client.go
```diff
@@ -43,6 +44,13 @@ var (
 	transport http.RoundTripper
 )
 
+// AMPEnabledForCountry reports whether the AMP transport should be wired up for
+// the given country. AMP fronts through Google domains that are unreachable
+// from China, so racing it there only wastes connection attempts.
+func AMPEnabledForCountry(country string) bool {
+	return !strings.EqualFold(country, "CN")
+}
+
```
garmr-ulfr: It might be a good idea to just have a function to enable/disable any transport. That sets us up to implement toggling transports from the config response in the future.
```go
func `EnableTransport(transport kindling.TransportName, enable bool) {
   EnabledTransport[transport] = enable
}
```

This also allows the caller to decide how to configure kindling and move this logic to `LocalBackend`;  kindling just needs to know what transports to use.

garmr-ulfr (approved):
LGTM!

PR #565, opened 2026-07-15
change the memory monitor dump path to the correct dir

garmr-ulfr:
This pull request includes a small but important change to how the memory dump file is handled in the `baseIssueAttachments` function. The memory dump is now sourced from the `logPath` directory instead of the `dataPath` directory, ensuring that the correct file location is used when attaching logs for issues. 

* Changed the source directory for `internal.MemoryDumpFileName` from `dataPath` to `logPath` in the `baseIssueAttachments` function in `backend/radiance.go`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Issue reports now attach memory dump files from the configured log directory, improving the accuracy of diagnostic attachments.


garmr-ulfr:
Bringing this in, just corrects a file path.

PR #580, opened 2026-07-27
smart: fix reversed disorder/split ordering, verify the dialer on Android

myleshorton:
Android logs showed the smart dialer's `disorder` strategy failing with `expected base dialer to return TCPConn`. Investigating turned up **two independent problems**, only one of which is Android-specific.

## 1. The strategy string is spelled backwards (all platforms)

`configurl` parses a pipeline left to right, so the **rightmost element is outermost**. `split:200|disorder:1` builds `disorder` on top of `split` — but `split.DialStream` returns a `transport.WrapConn`, and `disorder.DialStream` type-asserts its inner conn to `*net.TCPConn`. It can never succeed.

```mermaid
sequenceDiagram
    autonumber
    participant F as strategy finder<br/>outline-sdk smart
    participant D as disorder:1<br/>x/disorder/stream_dialer.go
    participant S as split:200<br/>transport/split/stream_dialer.go
    participant B as base dialer<br/>bypass/streamdialer.go

    Note over F: smart_dialer_config.yml<br/>entry spelled split:200 → disorder:1 ⚠️
    F->>D: DialStream — disorder is outermost<br/>configurl pipes left to right
    D->>S: DialStream on its inner dialer
    S->>B: DialStream
    B-->>S: *net.TCPConn
    Note over S: split wraps it in transport.WrapConn
    S-->>D: transport.StreamConn, no longer a *net.TCPConn
    rect rgba(255, 200, 200, 0.3)
        Note over D: innerConn.(*net.TCPConn) fails 🐛<br/>expected base dialer to return TCPConn
    end
    D-->>F: error — strategy silently dropped from the race
```

This failed with a plain `*transport.TCPDialer` too — nothing to do with the bypass dialer, and nothing to do with Android. Because `raceTests` just drops a losing strategy, we have been quietly running **four** of our five TLS strategies. outline-sdk's own docs use the other order (`disorder:0|split:123`).

Fixed by reordering to `disorder:1|split:200`.

## 2. `disorder` genuinely cannot run while the tunnel is up

`bypass.StreamDialer()` deliberately masks the loopback socket (`bypass/streamdialer.go`) whenever the SOCKS bypass proxy is listening. That masking is **correct** — setting TTL=1 on a `127.0.0.1` socket circumvents nothing, and unmasking it would let disorder win the search while doing no work. But it does mean disorder is unavailable on any connected session, on every platform (`vpn/boxoptions_default.go` adds the bypass inbound unconditionally). Android is just where it got noticed.

This is left as-is and asserted in a test, so removing the mask breaks the build.

## Tests

`kindling/smart/dialerconfig_test.go` — hermetic, runs on every platform:

- strict YAML parse against outline-sdk's schema (an unknown key is silently dropped at runtime today)
- no `system:` / `udp:` / `tcp:` resolver entries — the reason we fork kindling's config
- the disorder-ordering invariant, as a static check
- every strategy completing a real TLS handshake across **all three dialer shapes**: plain TCP, bypass-direct (pre-tunnel), bypass-proxied (connected). The proxied case asserts disorder fails *specifically* on the masked socket.

`kindling/smart/network_test.go` — opt-in via `RADIANCE_SMART_NETWORK_TEST=1`:

- each strategy against a real host individually, so one that always loses the race is still exercised — loopback can't test disorder, which needs a TTL-limited packet actually dropped in transit
- real config fetches in both dialer shapes, logging which strategy the search picked

I confirmed the tests catch the original bug by reverting the config: both the static ordering check and the TCP-dialer handshake fail.

## Running on Android

```
make test-android                                  # hermetic only
RADIANCE_SMART_NETWORK_TEST=1 make test-android    # + real egress
```

`scripts/android-test.sh` detects the device ABI, cross-compiles with the NDK, pushes, and runs. Three things it has to do that `gomobile bind` handles for an AAR but a bare test binary doesn't:

- `-ldflags=-checklinkname=0` — sing-box's `experimental/libbox` pull-linknames `os.checkPidfdOnce` to disable pidfd on Android, which the linker has rejected by default since Go 1.23. This is not a Go 1.26 regression: I reproduced the identical failure on 1.23, 1.24, 1.25 and 1.26, and sing-box's `go.mod` requires `go >= 1.23.1`, so no supported Go avoids it. `lantern-client`'s `gomobile bind` already passes the same flag.
- `CXX` pointed at the NDK — Go derives the default from the *target* GOOS and picks `g++`, which chokes on the Android objects (the C++ comes from `go-libutp`)
- pushing `libc++_shared.so` with `LD_LIBRARY_PATH`

`internal/testsocks` is a minimal SOCKS5 CONNECT fixture used by the smart-dialer tests, to keep the whole sing-box stack out of that test binary — 16 MB vs 67 MB, pushed to the device on every run. `bypass`'s own tests are unchanged and still use the real sing-box mixed inbound, including on Android.

## Verification

All tests pass locally and on an emulator (API 30, arm64-v8a), in both packages, including the real-network ones. `disorder:1|split:200` now completes a real TLS handshake from Android — its ~0.3 s vs ~0.04 s for the other strategies is the retransmit wait from the TTL-limited packet actually being dropped.

## Follow-up (not in this PR)

`getlantern/kindling`'s own `smart_dialer_config.yml` has the same reversed ordering. radiance overrides it so we're unaffected, but any other `NewSmartHTTPTransport` caller is running a dead strategy. Separate one-line PR to follow.

## Summary by CodeRabbit

* **New Features**
  * Added Android test support for connected devices and emulators.
  * Added automated smart-network integration tests for direct and bypass-proxy connections.
  * Added SOCKS5 proxy support for testing network flows.

* **Bug Fixes**
  * Updated smart dialing strategy ordering to improve connection reliability.

* **Documentation**
  * Clarified smart dialer configuration requirements and strategy behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,3 +65,9 @@ proto:
 .PHONY: test
 test:
 	go test -v ./...
+
+# Runs on a connected device or emulator. Set RADIANCE_SMART_NETWORK_TEST=1 to
+# include the tests that need real egress.
+.PHONY: test-android
+test-android:
+	./scripts/android-test.sh
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/getlantern/dnstt v0.0.0-20260603191204-3b860502c0ac
 	github.com/getlantern/domainfront v0.0.0-20260722204513-8c1f8acfa715
 	github.com/getlantern/keepcurrent v0.0.0-20260616120552-f204338b01a3
-	github.com/getlantern/kindling v0.0.0-20260722205227-8563bcdd9b04
+	github.com/getlantern/kindling v0.0.0-20260727200630-e0ccaa2a6abd
 	github.com/getlantern/lantern-box v0.0.106
 	github.com/getlantern/pluriconfig v0.0.0-20251126214241-8cc8bc561535
 	github.com/getlantern/publicip v0.0.0-20260328175246-2c460fe80c6b
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ replace github.com/refraction-networking/water => github.com/getlantern/water v0
 require (
 	github.com/1Password/srp v0.2.0
 	github.com/Jigsaw-Code/outline-sdk v0.0.19
+	github.com/Jigsaw-Code/outline-sdk/x v0.0.2
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/alexflint/go-arg v1.6.1
 	github.com/alitto/pond v1.9.2
@@ -64,7 +65,6 @@ require (
 
 require (
 	filippo.io/edwards25519 v1.1.0 // indirect
-	github.com/Jigsaw-Code/outline-sdk/x v0.0.2 // indirect
 	github.com/RoaringBitmap/roaring v1.2.3 // indirect
 	github.com/STARRY-S/zip v0.2.3 // indirect
 	github.com/akutz/memconn v0.1.0 // indirect
@@ -302,7 +302,7 @@ require (
 	golang.zx2c4.com/wintun v0.0.0-20230126152724-0fa3db229ce2 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260504160031-60b97b32f348 // indirect
 	google.golang.org/grpc v1.80.0
-	gopkg.in/yaml.v3 v3.0.1 // indirect
+	gopkg.in/yaml.v3 v3.0.1
 	lukechampine.com/blake3 v1.3.0 // indirect
 )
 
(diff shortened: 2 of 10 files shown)


myleshorton:
kindling#45 is merged (`e0ccaa2`), so I've pushed a follow-up commit bumping the pin to it.

That commit is exactly the same reordering, applied upstream — it's the *only* commit between the previous pin and kindling main, and radiance overrides the embedded config regardless, so the bump is behaviourally inert here.

What it does change is the resync story. radiance's copy now differs from kindling's by exactly one line:

```
2d1
<   - system: {}
```

so I've narrowed the `DialerConfig` doc back to that single divergence rather than the two it described while the pin was stale.

Re-verified after the bump: full suite green locally and on the emulator (API 30, arm64-v8a), including the opt-in real-egress tests.

myleshorton:
Thanks both — went through all four findings. Two fixed, two I'm rejecting with evidence.

**Fixed**

- `network_test.go` env gate (Copilot): correct, `RADIANCE_SMART_NETWORK_TEST=0` enabled the tests. Now parsed with `strconv.ParseBool`, so ``/`0`/`false` skip and `1`/`true` run. Verified across all five values.
- `testsocks.Server`/`Addr` (CodeRabbit): neither had a caller. Rather than add doc comments that restate the identifier names — which this repo's Code Comments guidance explicitly rules out — I deleted the exported surface. `Listen` is now the whole API.

**Not fixed — the premises don't hold on macOS**

- `sort -V` (Copilot, `android-test.sh:46`): macOS `sort` is FreeBSD-derived and does support `-V`. On macOS 26.5.1:
  ```
  $ printf '29.0.1\n9.0.0\n26.1.2\n' | /usr/bin/sort -V
  9.0.0
  26.1.2
  29.0.1
  ```
  No coreutils on PATH. `ANDROID_NDK_HOME` already overrides detection if a host ever does disagree.
- `mktemp -d` (Copilot, `android-test.sh:93`): works without a template on macOS — that's where the build dirs in every run of this script came from, e.g. `/var/folders/tb/.../tmp.dOwF5ocKaP`. Directly:
  ```
  $ /usr/bin/mktemp -d
  /var/folders/tb/.../tmp.iyXCAEovYA
  ```

Adding fallbacks for either would be complexity guarding a condition I can't reproduce. Happy to add them if someone hits it on an older macOS.

Full suite green locally after the changes.

PR #584, opened 2026-07-29
remove name from tun options to avoid possible conflicts

garmr-ulfr:
This pull request makes a small change to the `vpn/boxoptions_default.go` file by removing the explicit setting of the `InterfaceName` field in the `TunInboundOptions` struct. This allows the OS to assign one, avoiding possible conflicts.

closes getlantern/engineering/issues/3707.


garmr-ulfr:
Bringing this in, we don't have any reason/need to fix the TUN name.

PR #599, opened 2026-08-13
backend: cap the servers a client holds at 30

myleshorton:
Closes getlantern/engineering#3816.

## The gap

`maxRetainedLanternServers` was applied only to *retention*:

```go
retentionBudget := max(limit-incomingCount, 0)
```

That budgets how many **existing** servers to keep. It never truncated the **incoming** batch, so a config delivering N servers added all N and simply evicted whatever was retained. The limit was 60; production configs deliver **120**, measured on two independent devices:

| ticket | device | `servers.json` entries |
|---|---|---|
| 181331 | Derek, iOS | 120 |
| 181479 | Adam, iOS | 120 |

So clients held double the nominal cap, with no client-side bound at all.

## Why it matters

The iOS network extension runs under a **fatal 50 MB** jetsam cap, and startup work scales with this set — each outbound carries adapter state, the offline URL tests ran with `server_count=60`, and the smart dialer probes `11 DNS entries x 5 TLS strategies` per config host.

Even after #596 and #597, a device capture (iPhone 16 / iOS 26.6) peaked at **46.08 MB — 3.92 MB from the cap** — while cycling connect/disconnect:

```
Tunnel[1263]  40.53 MB  (avail 9.47)
Tunnel[1263]  44.66 MB  (avail 5.34)
Tunnel[1263]  46.08 MB  (avail 3.92)
Tunnel[1281]  30.74 MB  (fresh process, healthy)
```

Five distinct PIDs, so not a leak across restarts — the growth is within one long-lived extension, and it mirrors the churn in the sessions that were being killed.

## The change

Cap at **30**, enforced on the incoming batch as well as on retention, so no config can push a client past it. Config order is the server's preference, so the leading entries are kept; the existing quality ordering (hard-demoted first, then selection age) still decides which retained servers go. Private servers are not Lantern servers and are unaffected.

This is a client-side guard — the server can still offer more, and the client keeps the best subset.

## Tests

`TestClientHoldsNoMoreThanTheCapAfterAnUpdate` asserts the invariant that actually broke: the **total** a client is left holding after an update, not just what it retained. Verified it fails without the fix, reproducing the production number exactly:

```
client would hold 120 servers, over the 30 cap
```

`go build -tags with_clash_api ./...`, `go vet`, and `go test -race ./backend/` all clean.

## Worth a look before merge

30 is what we agreed as the target. It is a behavioural change for users who currently get 120 — fewer servers to fail over to, so it is worth a sanity check that the bandit still has enough choice in hostile regions, where more outbounds may matter more than memory. Happy to make it platform-conditional (tight on iOS, looser elsewhere) if that trade reads better.

## Summary by CodeRabbit

- **Bug Fixes**
  - Limited clients to a maximum of 30 configured servers.
  - Ensured oversized server configurations retain the preferred entries within the limit.
  - Preserved private server availability without applying the cap.

- **Tests**
  - Added coverage for oversized, exact-limit, undersized, and empty server configurations.

### DIFF
--- a/backend/radiance.go
+++ b/backend/radiance.go
@@ -766,9 +766,13 @@ func (r *LocalBackend) RevokePrivateServerInvite(ip string, port int, accessToke
 	return r.srvManager.RevokePrivateServerInvite(ip, port, accessToken, inviteName)
 }
 
-// maxRetainedLanternServers caps the number of working Lantern servers retained
-// across config updates.
-const maxRetainedLanternServers = 60
+// maxLanternServers caps how many Lantern servers a client holds. The iOS network
+// extension runs under a fatal 50 MB jetsam cap and its startup work scales with
+// this set — each outbound carries adapter state, and the offline URL tests and
+// strategy search run across it. Enforced on the incoming batch as well as on
+// retention, so no config can push a client past it. Private servers are not
+// Lantern servers and are unaffected.
+const maxLanternServers = 30
 
 func (r *LocalBackend) updateServers(list servers.ServerList) error {
 	existing := r.srvManager.AllServers()
@@ -778,7 +782,16 @@ func (r *LocalBackend) updateServers(list servers.ServerList) error {
 		return exists
 	})
 
-	tagsToEvict := lanternServersToEvict(existing, len(list.Servers), maxRetainedLanternServers)
+	if capped := capServerBatch(list.Servers, maxLanternServers); len(capped) < len(list.Servers) {
+		slog.Debug(
+			"Truncating config batch to the client outbound cap",
+			"received", len(list.Servers),
+			"cap", maxLanternServers,
+		)
+		list.Servers = capped
+	}
+
+	tagsToEvict := lanternServersToEvict(existing, len(list.Servers), maxLanternServers)
 
 	if len(tagsToEvict) > 0 {
 		slog.Debug(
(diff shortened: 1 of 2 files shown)


myleshorton:
Closing — not taking this now. Recording what it found so the work is not lost.

The defect it uncovered is real and still open in engineering#3816: `maxRetainedLanternServers` is applied only to retention —

```go
retentionBudget := max(limit-incomingCount, 0)
```

— so it budgets how many *existing* servers to keep and never truncates the incoming batch. The limit was 60; production configs deliver **120**, measured on two independent devices (tickets 181331 and 181479). Clients hold double the nominal cap with no client-side bound at all.

That remains true after everything shipped in v9.1.23-beta. Peak under connect/disconnect cycling was still 44.33 MB before the probe optimization and 38.88 MB after, against a fatal 50 MB cap.

Reason for holding: capping at 30 is a behavioural change, not just a memory optimization — fewer servers to fail over to, and in hostile regions more outbounds may matter more than memory. iOS is also the only platform with the 50 MB constraint, so a flat cap taxes Android and desktop for no benefit. If this is revisited, making it platform-conditional (tight on iOS, looser elsewhere) is probably the better shape.

The branch `fisk/cap-client-outbounds` keeps the implementation and `TestClientHoldsNoMoreThanTheCapAfterAnUpdate`, which was verified to fail without the fix (reproducing the production number: `client would hold 120 servers, over the 30 cap`).

PR #601, opened 2026-08-14
issue: bound the bytes the archiver holds in memory

myleshorton:
Submitting an issue report from the iOS app **while the VPN is connected** killed the tunnel. Closes the immediate crash in getlantern/engineering#3820.

## What was happening

The extension was healthy one second before the submit — this is the archiver, not tunnel memory growth:

```
16:43:15.578  extension healthy — footprint 31.25 MB, available 18.75 MB
16:43:16.142  POST /issue arrives
              ...per-second memory samples stop entirely...
16:43:16      app: Error reporting issue: ipc request POST /issue failed
16:43:21      Disconnecting / Disconnected      <-- unprompted
16:43:22      retry succeeds (tunnel down, so the *app* builds the archive)
```

Reproduced twice on iPhone 16 / iOS 26.6.

## Why

`buildIssueArchive` reads whole files into memory and zips into a `bytes.Buffer`. `maxLogReadFactor` looks like a bound but is a **compression** bound, not a memory one — `maxArchiveSize * 20` is **390 MB per file** — and it was applied per file rather than across the set. The reproducing device had **42 MB** in the log dir:

| file | size |
|---|---|
| `lantern.log` | 26 MB |
| `lantern_ios.log` | 10 MB |
| attachments | ~68 KB |

42 MB of reads plus the zip buffer, on top of a ~31 MB working set, inside a process with a **fatal 50 MB** jetsam cap.

## The change

One budget threaded through every read, so the peak is bounded by the platform rather than by whatever happens to be on disk, capped at **8 MB on mobile**. Reading the tail means a smaller budget costs history, not correctness. `prependMostRecentBackup` now draws from the same pool, since it decompresses into memory too.

`snapshotLogFile`'s second parameter is now bytes rather than a compressed-size budget to scale up; its existing test moves with it.

## Scope — deliberately partial

This **bounds** the archiver. It does not stop the archive being built inside the extension, which is the better fix: `IssueReporter` only needs an HTTP client and `settings.LogPathKey`, and on iOS that log dir is the **shared App Group container** the app can read directly. Moving report-building to the app removes the 50 MB constraint entirely. Tracked in engineering#3820, and worth doing next — this is the stop-the-bleeding half.

Streaming the archive to a temp file instead of buffering is the other half. It is a real refactor (`fitArchive` buffers deliberately, because it needs the *compressed* size to decide how much of the primary log to keep, and `issue.go` then `proto.Marshal`s the bytes, duplicating them again), so it wants its own PR rather than being rushed in alongside this.

## Tests

`TestArchiveInputBudgetIsBoundedOnMobile` asserts the budget is actually bounded on mobile and documents how loose the compression-derived one is (>300 MB). The updated `TestSnapshotLogFile` covers the byte-cap semantics.

I wrote and then **deleted** a `TestBuildIssueArchiveStopsReadingAtTheBudget` — it passed with and without the fix, because it measured the zip output (already truncated by `fitArchive`) rather than bytes read into memory. Peak read memory is not observable from the archive output, so a test asserting it there is theatre. Better to say so than ship a guard that does not guard.

`go build -tags with_clash_api`, `gofmt`, and `go test -race ./issue/` all clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Log archives now use a shared input limit across snapshots, attachments, and rotated backups.
  - Mobile devices limit archived log data to 8 MiB, helping prevent excessive resource usage.
  - When the limit is reached, additional logs are skipped.
  - Rotated backups now receive only the remaining available archive capacity.
  - Desktop archiving continues to respect its existing compression-based limits.

### DIFF
--- a/issue/archive.go
+++ b/issue/archive.go
@@ -11,14 +11,23 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/getlantern/radiance/common"
 )
 
 const (
 	// maxLogReadFactor bounds how much uncompressed log to read toward an archive:
 	// maxArchiveSize * maxLogReadFactor bytes. Logs compress by at most roughly this
 	// factor, so reading more than this can never be needed to reach the compressed
-	// size budget.
+	// size budget. That is a compression bound, not a memory one — see
+	// maxMobileInputBytes.
 	maxLogReadFactor int64 = 20
+	// maxMobileInputBytes caps the uncompressed bytes the archiver holds at once on
+	// mobile. maxLogReadFactor alone permits 19.5 MB * 20 = 390 MB per file, no
+	// constraint at all inside the iOS network extension's fatal 50 MB jetsam cap:
+	// reporting an issue while connected read 42 MB of logs and the extension was
+	// killed mid-archive, dropping the tunnel. engineering#3820.
+	maxMobileInputBytes int64 = 8 * 1024 * 1024
 	// backupTimeFormat matches the timestamp format used in rotated backup
 	// filenames: "<log-name>-<timestamp>.log.gz".
 	backupTimeFormat = "2006-01-02T15-04-05.000"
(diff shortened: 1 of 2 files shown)


myleshorton:
Closing — superseded in practice by #603, which assembles mobile issue reports in the calling process. The crash this was written to stop is fixed at the better layer: the app now builds the archive, where the log dir is equally reachable through the shared App Group container and the ceiling is not 50 MB. Confirmed on device — submitting a report while connected now succeeds and the tunnel stays up.

Two things this PR found that #603 does **not** address, noted on engineering#3820 so they are not lost:

1. `maxLogReadFactor` reads as a bound but is a *compression* bound — `maxArchiveSize * 20` is 390 MB per file — and it is applied per file rather than across the set. The archiver is still unbounded in principle; it just no longer runs in the process with the hard ceiling.
2. The archive is still fully buffered: `os.ReadFile` per file into a `bytes.Buffer` zip, and `issue.go` then `proto.Marshal`s the bytes, duplicating them again. On the reproducing device that is 42 MB read and held twice. Streaming to a temp file remains the real fix, and `fitArchive` buffers deliberately because it needs the compressed size, so it is a genuine refactor rather than a patch.

Neither is urgent now that the 50 MB constraint is out of the path. Branch `fisk/bound-issue-archive-memory` keeps the implementation if it is wanted later.

PR #613, opened 2026-08-20
Add user-message client subsystem

atavism:
Adds the Radiance-side support for user messaging.

## New Features

* Added personalized in-app messages based on account, locale, platform, and activity.
* Messages can be retrieved, refreshed, acknowledged, and updated as account or activity status changes.
* Added reliable polling, retry handling, and safeguards against repeated displays.
* Message state persists across restarts, while expired or invalid messages are safely discarded.

## Bug Fixes

* Improved handling of unsupported, malformed, or unavailable messages.
* Sensitive request and configuration details are now excluded from diagnostic logs.


usermessage/service.go
```diff
@@ -0,0 +1,306 @@
...
+		requestContext, requestID, generation, ok := s.beginRequest(ctx)
+		if !ok {
+			delay = 0
+			continue
+		}
+		clientContext := s.contextProvider()
+		seen := s.store.seen(clientContext.UserID)
+		response, err := s.fetcher.Fetch(requestContext, clientContext, seen)
+		s.endRequest(requestID)
+		if errors.Is(err, context.Canceled) {
```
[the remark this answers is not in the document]
atavism: Fixed in 5a7b0f1. An incomplete client context is now treated as a deferred state: the request is ended, failure state is reset, and the service blocks until an account/context refresh instead of entering the retry loop. TestServiceWaitsForCompleteCredentials covers the behavior.

usermessage/service.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	return delay
+}
+
+func defaultJitter(delay time.Duration) time.Duration {
+	if delay <= 0 {
+		return 0
+	}
+	return time.Duration(float64(delay) * (0.9 + rand.Float64()*0.1))
+}
```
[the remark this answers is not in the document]
atavism: Verified against the pinned Go 1.26 toolchain; no code change is needed here. Package-level math/rand has been randomly seeded at process startup since Go 1.20 (and rand.Seed is a no-op as of Go 1.24), so Float64 does not repeat the same sequence across process starts. The existing bounds test still covers the SLA.

usermessage/service.go
```diff
@@ -0,0 +1,306 @@
...
+	var failures uint
+	for s.wait(ctx, delay) {
+		requestContext, requestID, generation, ok := s.beginRequest(ctx)
+		if !ok {
+			delay = 0
+			continue
+		}
+		clientContext := s.contextProvider()
+		seen := s.store.seen(clientContext.UserID)
+		response, err := s.fetcher.Fetch(requestContext, clientContext, seen)
```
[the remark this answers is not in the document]
atavism: Fixed in 5a7b0f1 with the same deferred-context path. I did not use the suggested delay=0 loop because that would immediately re-enter polling while the service is active; the new path waits on the refresh signal and avoids both retries and a busy loop.

usermessage/http.go
```diff
@@ -0,0 +1,127 @@
...
+	ProToken   string
+	Locale     string
+	Platform   string
+	AppVersion string
+}
+
+func (c ClientContext) valid() bool {
+	userID, err := strconv.ParseUint(c.UserID, 10, 64)
+	return err == nil && userID > 0 && c.ProToken != "" && len(c.ProToken) <= 4096 &&
+		c.Locale != "" && c.Platform != "" && c.AppVersion != ""
```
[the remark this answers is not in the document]
atavism: Fixed in 5a7b0f1. Client validation now parses the same signed 64-bit range as Lantern Cloud and requires an exact FormatInt round trip. Tests cover leading zeros, a leading plus, overflow, and non-numeric IDs.

backend/radiance.go
```diff
@@ -1604,7 +1666,11 @@ func (r *LocalBackend) CompleteRecoveryByEmail(ctx context.Context, email, newPa
...
 func (r *LocalBackend) DeleteAccount(ctx context.Context, email, password string) (*account.UserData, error) {
-	return r.accountClient.DeleteAccount(ctx, email, password)
+	userData, err := r.accountClient.DeleteAccount(ctx, email, password)
+	if err == nil {
+		r.refreshUserMessages()
+	}
+	return userData, err
 }
 
 func (r *LocalBackend) SignUp(ctx context.Context, email, password string) ([]byte, *account.SignupResponse, error) {
```
garmr-ulfr, line 1685: Do we need to refresh user messages here too?

backend/radiance.go
```diff
@@ -103,8 +106,9 @@ type Options struct {
 	// own device ID and ignore this value
 	DeviceID string
 	// User choice for telemetry consent
-	TelemetryConsent  bool
-	PlatformInterface vpn.PlatformInterface
+	TelemetryConsent        bool
+	PlatformInterface       vpn.PlatformInterface
+	UserMessageCapabilities wire.ClientCapabilities
```
garmr-ulfr, line 111: Since this is passed in through the options, we'll need to update the [IPC client for mobile](https://github.com/getlantern/radiance/blob/main/ipc/client_mobile.go#L24-L29) to store it and pass it through when [re-creating the fallback backend](https://github.com/getlantern/radiance/blob/main/ipc/client_mobile.go#L24-L29) so it works when then tunnel isn't up.

usermessage/service.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+type realClock struct{}
+
+func (realClock) Now() time.Time                 { return time.Now() }
+func (realClock) NewTimer(d time.Duration) Timer { return realTimer{time.NewTimer(d)} }
+
+type realTimer struct{ *time.Timer }
+
+func (t realTimer) C() <-chan time.Time { return t.Timer.C }
```
garmr-ulfr: Can we update the tests to use `require.Eventually`/`Never` so we don't need to define clock/timer interfaces? We could use small intervals in tests to avoid long waits.

usermessage/service.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+	return delay
+}
+
+func defaultJitter(delay time.Duration) time.Duration {
+	if delay <= 0 {
+		return 0
+	}
+	return time.Duration(float64(delay) * (0.9 + rand.Float64()*0.1))
+}
```
garmr-ulfr, line 438: `common.Backoff` already implements backoff with jitter. I think expanding it support waiting on a chan as well would be cleaner than re-implementing it.
```go
func (b *Backoff) Wait(ctx context.Context) {
	b.WaitOn(ctx, nil)
}

func (b *Backoff) WaitOn(ctx context.Context, wake <-chan struct{}) {
	...

	select {
	case <-ctx.Done():
	case <-wake:
	case <-time.After(wait):
	}
}
```

usermessage/service.go
```diff
@@ -0,0 +1,420 @@
...
+			s.logger.Warn(
+				"User-message fetch result could not be persisted",
+				"category", "local_state",
+				"failure_count", failures,
+				"retry_in", delay,
+			)
+			continue
+		}
+		failures = 0
+		delay = s.jitter(time.Duration(response.PollIntervalSeconds) * time.Second)
```
garmr-ulfr: On a successful response, I think the client should respect `response.PollIntervalSeconds` exactly. The server can return a jittered delay if it's needed.

usermessage/service.go
```diff
@@ -0,0 +1,420 @@
...
+			// Account creation normally wakes us through Refresh, but first-run
+			// identity creation has historically not emitted that signal. Recheck
+			// the in-memory context so a missed edge cannot stall messaging until
+			// the next app launch. No network request is made until it is valid.
+			delay = credentialRecheckInterval
+			if !credentialsDeferred {
+				s.logger.Debug("User-message fetch deferred", "reason", "credentials_unavailable")
+				credentialsDeferred = true
+			}
+			continue
```
garmr-ulfr: Right now an incomplete context (id but no token, say) rechecks every second for as long as the app's active. Could we return `errCredentialsUnavailable` and let it ride the same failure backoff as everything else? That'd drop the special recheck branch and the `credentialsDeferred` flag, and cap the idle wakeups at the backoff max instead of 1 second.

usermessage/store.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	if err != nil {
+		return nil, fmt.Errorf("read user-message state: %w", err)
+	}
+	if err := json.Unmarshal(data, &s.state); err != nil {
+		return nil, fmt.Errorf("decode user-message state: %w", err)
+	}
+	if s.state.Version != stateVersion {
+		return nil, fmt.Errorf("unsupported user-message state version %d", s.state.Version)
+	}
```
garmr-ulfr, line 68: A corrupt or unsupported `user-messages.json` currently disables user messages for the session, and because the file remains in place it may keep disabling the feature on every launch. Worst case, mobile users may have to reinstall the app just to recover. Could we quarantine/reset invalid user-message state and continue with an empty store?

usermessage/service.go
```diff
@@ -0,0 +1,420 @@
...
+			failures++
+			delay = s.jitter(failureBackoff(failures))
+			s.logger.Warn(
+				"User-message fetch result could not be persisted",
+				"category", "local_state",
+				"failure_count", failures,
+				"retry_in", delay,
+			)
+			continue
+		}
```
garmr-ulfr: I don’t think file writes should be treated like a fetch failure and retried with the network backoff. The fetch already succeeded; the local state is unhealthy. If writing to the file failed once because of permissions, disk, path, or filesystem state, repeated fetches are likely to keep failing.

usermessage/store.go
```diff
@@ -0,0 +1,267 @@
...
+	if state == nil {
+		state = &userState{}
+		next.Users[userID] = state
+		touch(&next, userID)
+	}
+	if slices.Contains(state.Seen, message.DisplayID) || state.Pending != nil {
+		return nil
+	}
+	state.Pending = cloneMessage(message)
```
garmr-ulfr, line 150: FYI, `touch` only fires for new users and on `acknowledge`, so an existing account that keeps getting messages but never acks won't get bumped, and could be evicted ahead of quieter ones once there are ≥16 on a device.

usermessage/service.go
```diff
@@ -0,0 +1,420 @@
...
+	jitter          func(time.Duration) time.Duration
+	logger          *slog.Logger
+	store           *store
+	wake            chan struct{}
+
+	mu            sync.Mutex
+	started       bool
+	active        bool
+	online        bool
+	generation    uint64
```
garmr-ulfr: What is ~`generation`~ `refreshGeneration` used for?

atavism:
I took a look at the fake-clock and backoff suggestions. I think we should keep the current approach for now. The fake clock keeps polling, retry, cancellation, and wake tests deterministic, while `common.Backoff` doesn’t support the timing or wake behavior this service needs. We can extract a shared helper later if other subsystems need the same thing.

garmr-ulfr:
> I took a look at the fake-clock and backoff suggestions. I think we should keep the current approach for now. The fake clock keeps polling, retry, cancellation, and wake tests deterministic, while `common.Backoff` doesn’t support the timing or wake behavior this service needs. We can extract a shared helper later if other subsystems need the same thing.

I think both concerns can be addressed without the clock abstractions: the wake behavior can be added to the shared backoff helper, and the retry schedule can be tested as a pure delay calculation.

**Wake.** Expanding `common.Backoff` with the wake-aware variant adds that support:

```go
func (b *Backoff) Wait(ctx context.Context) {
	b.WaitOn(ctx, nil)
}

func (b *Backoff) WaitOn(ctx context.Context, wake <-chan struct{}) {
	...

	select {
	case <-ctx.Done():
	case <-wake:
	case <-time.After(wait):
	}
}
```

A nil `wake` is never ready, so `Wait(ctx)` keeps the same behavior for existing callers.

**Timing.** The clock exists so the service test can assert the retry schedule from outside the loop. If the ladder lives in `common.Backoff` instead, split the computation off the sleep:

```go
func (b *Backoff) WaitOn(ctx context.Context, wake <-chan struct{}) {
	if ctx.Err() != nil {
		return
	}

	wait := b.nextDelay(rand.Float64())
	select {
	case <-ctx.Done():
	case <-wake:
	case <-time.After(wait):
	}
}

func (b *Backoff) nextDelay(random float64) time.Duration {
	b.n++
	wait := b.baseWait * time.Duration(b.n*b.n)
	jitter := 0.8 + 0.4*random
	return min(b.maxWait, time.Duration(float64(wait)*jitter))
}
```

Now the schedule is assertable with exact equality, no sleeping and no clock:

```go
backoff := NewBackoff(time.Second, time.Minute)
for _, expected := range []time.Duration{1 * time.Second, 4 * time.Second, 9 * time.Second} {
	require.Equal(t, expected, backoff.nextDelay(0.5))
}
backoff.Reset()
require.Equal(t, time.Second, backoff.nextDelay(0.5))

require.Equal(t, 800*time.Millisecond, NewBackoff(time.Second, time.Minute).nextDelay(0))
require.Equal(t, time.Minute, NewBackoff(time.Minute, time.Minute).nextDelay(1))
```

That's the same shape as `TestFailureJitterAndBackoffCap` in this PR, which already tests the curve as a pure function with no clock — it would just move to `common`, which has no tests at all right now (no test file in the repo references `common.Backoff`) despite four callers: `config`, `ipc`, `account/datacap`, and `lanternd`.

With the curve covered there, the service test only has to prove it delegates — `WaitOn` on failure, `Reset` on success — which is a double, not a clock:

```go
// service.go

type failureBackoff interface {
	WaitOn(ctx context.Context, wake <-chan struct{})
	Reset()
}

var newFailureBackoff = func() failureBackoff {
	return common.NewBackoff(initialFailureBackoff, maxFailureBackoff)
}

// service_test.go

func useFailureBackoff(t *testing.T, backoff failureBackoff) {
	t.Helper()
	previous := newFailureBackoff
	newFailureBackoff = func() failureBackoff { return backoff }
	t.Cleanup(func() { newFailureBackoff = previous })
}

type recordingBackoff struct{ 
	calls chan string
}

func (b *recordingBackoff) WaitOn(context.Context, <-chan struct{}) { b.calls <- "wait" }
func (b *recordingBackoff) Reset()                                  { b.calls <- "reset" }
```

Then the test asserts `wait`, `wait`, `reset`, `wait` over the same fetch handoffs it uses now. I tried this locally: dropping either `Reset()` or the wait from the loop fails it, and dropping the `n*n` or the `maxWait` clamp fails the `common` tests.

For the successful poll interval, return the delay from the fetch and assert the returned value — `require.Equal(t, 5*time.Minute, delay)` for `PollIntervalSeconds: 300` — rather than observing the timer the loop creates.

`Clock`, `Timer`, `realClock`, `realTimer`, `fakeClock`, and `fakeTimer` all go away, and `Options` stops exporting `Clock` and `Jitter` — those are a test seam on our public API. Worth noting the fake clock isn't what makes the cancellation and wake tests deterministic today: `TestServiceStopsAfterParentContextCancellation` constructs a clock and never advances or reads it. That determinism comes from the fetcher handoff, which stays either way.

`common.Backoff` is quadratic, which is smoother than exponential, so it reaches the 5m cap on the 8th failure instead of the 7th and is sparser early. But, if you feel strongly about using exponential instead, I'd rather change `common.Backoff` to exponential than keep a second implementation.

The fake clock also serves `Now()` for the message-expiry assertions; those would use a short real TTL with `require.Eventually`, which is how expiry is tested elsewhere.

usermessage/service.go
garmr-ulfr: Would you add a comment explaining what `refreshGeneration` is used for?

garmr-ulfr:
Please reply to all comments, unless they genuinely don't warrant a response, instead of just marking them resolved. If a change was made to address a concern, say so and marked as resolved, otherwise say why it wasn't addressed and leave it open so the reviewer can respond or they'll mark as resolved. Couple you also add a PR description?

atavism:
@jigar-f Can you let me know if this one looks good to you?

jigar-f (approved):
LGTM from my side. I’ll have @garmr-ulfr handle the review of this. Thanks.

PR #644, opened 2026-09-16
Attach lantern-box connection diagnostics to user reports

myleshorton:
User-submitted connection reports currently lack socket evidence to help distinguish connection timeouts, resets, and transport failures. Enable lantern-box’s bounded local diagnostics and attach a snapshot as `attachments/connection-diagnostics.json` in the existing report ZIP, subject to its attachment size budget.

The collector retains at most 512 attempts, hashes endpoints and outbound tags with a process-local key, and excludes direct outbounds. Radiance enables local collection at startup. It serializes and uploads a snapshot only when constructing a user-submitted report. Platform collectors and history live in lantern-box; generic physical socket hooks live in sing-box-minimal.

Extend the ZIP helper to accept generated in-memory attachments and handle budgets smaller than ZIP overhead. Observed application I/O counters are partial; these diagnostics cannot independently prove DPI or IP blocking.

Depends on getlantern/sing-box-minimal#63 and getlantern/lantern-box#327, pinned to published commits. Merge the dependency PRs first.

Validation: focused race tests cover report attachments and size budgets; VPN/backend compilation is checked against the published dependencies. Collector platform builds cover Linux, Android, iOS, and Windows. Mobile and Windows device runtime testing remains outstanding.

Session history exposes `diagnostic_outbound_id` for joining selected sessions to socket records within the same process. This preserves endpoint obfuscation; it does not provide cross-restart correlation or identify never-selected probe outbounds. The collector supports disabling and clearing history through `connectiondiag.Disable()`; a user-facing setting is not part of this PR. Observed I/O is explicitly marked partial, and zero does not mean no traffic flowed.

Before merging: replace both dependency commit pins with release tags after their upstream PRs merge and releases are published.


go.sum
```diff
@@ -276,6 +278,10 @@ github.com/getlantern/semconv v0.0.0-20260327040646-21845dda05cb h1:c5YM7b3a4r2J
 github.com/getlantern/semconv v0.0.0-20260327040646-21845dda05cb/go.mod h1:GkPT5P9JoOTIRXRmFWxYgu1hhXgTFFTNc2hoG7WQc3g=
 github.com/getlantern/sing-box-minimal v1.13.19-lantern h1:dwVCskH9N3Q5a2wVELdyGDOHKvqWK0D9/rCM/VcaDF0=
 github.com/getlantern/sing-box-minimal v1.13.19-lantern/go.mod h1:4E8WvF7Vi60N1NtPw1CeG994J/yiJZZkty5dKKwG+V0=
+github.com/getlantern/sing-box-minimal v1.13.20-0.20260916171646-f42548ba8366 h1:vE6lRs4rWF8RSXrKV2daUSEzW6i3gEbU+O8k4KFKnWc=
```
myleshorton: `go mod tidy` was not run. It removes six stale lines: the old `lantern-box v0.0.130`, the old `sing-box-minimal v1.13.19-lantern`, and this intermediate `f42548ba8366` pseudo-version that nothing references anymore.

go.mod
```diff
@@ -45,7 +45,7 @@ require (
 	github.com/getlantern/domainfront v0.0.0-20260915172349-4984689a8667
 	github.com/getlantern/keepcurrent v0.0.0-20260616120552-f204338b01a3
 	github.com/getlantern/kindling v0.0.0-20260915174455-ef1078a2dd49
-	github.com/getlantern/lantern-box v0.0.130
+	github.com/getlantern/lantern-box v0.0.131-0.20260916180129-440b431fa495
```
myleshorton: Both this and the `replace` on line 7 pin PR-branch pseudo-versions. Prior convention was tags (`lantern-box v0.0.130`, `sing-box-minimal v1.13.19-lantern`). After getlantern/lantern-box#327 merges, tag it and re-pin here; same for sing-box-minimal (`v1.13.20-lantern`). If either upstream is squash-merged first, these commits become unreachable and fresh builds fail on `go mod download`.

common/reporting/reporting.go
```diff
@@ -6,11 +6,13 @@ import (
...
 	"runtime/debug"
 
+	"github.com/getlantern/lantern-box/connectiondiag"
 	"github.com/getlantern/radiance/log"
 )
 
-// Init records the build version in the log.
+// Init enables local connection diagnostics and logs the build version.
 func Init(version string) {
+	connectiondiag.Enable(false)
```
myleshorton, line 15: This makes collection always-on for every user from process start, whereas the lantern-box PR describes it as opt-in. The footprint is small (≤512 records, one atomic add per socket read), so I think always-on is the right default, but it should be a deliberate decision, and there is currently no `Disable()` to hang a settings flag on later. Also: on Android the observed byte counters will read ~0 because sing's splice path bypasses the wrapper (see the comment on getlantern/lantern-box#327), so anything that consumes this attachment shouldn't treat zero bytes as "no traffic".

myleshorton (commented):
Reviewed with getlantern/sing-box-minimal#63 and getlantern/lantern-box#327. `./issue` passes under `-race`; the ZIP helper correctly returns no archive for budgets below the 22-byte empty-ZIP minimum and `Report` already skips a nil archive. Two hygiene items inline plus one design note. Merge after lantern-box, then re-pin both deps to tags.

go.sum
myleshorton: Ran go mod tidy in be618ab after repinning both upstream review fixes. Removed stale intermediate sums. Focused issue/session race tests pass against the published dependencies, and backend compilation passes.

go.mod
myleshorton: Updated both pins to published review-fix commits in be618ab. The release-tag step remains open: merge/tag getlantern/sing-box-minimal#63, repin and merge/tag getlantern/lantern-box#327, then repin this PR to both release tags before merging. This dependency order is now explicit in all three PR descriptions.

common/reporting/reporting.go
myleshorton: Deliberately keeping bounded local collection enabled at Radiance startup, with upload only through a user bug report. The description now distinguishes collection from serialization/upload. Upstream 204410f adds Disable() to stop new collection and clear history without closing connections; a settings UI is outside this PR. Records now carry io_coverage="partial" and document that zero observed bytes cannot establish no traffic, including Android splice paths.

PR #646, opened 2026-09-18
Advertise CapabilityIPv6 in ConfigRequest when the device has real v6 reachability

AGMETEOR:
## Summary
- lantern-cloud gates `address_family='v6'` proxy tracks (routes with an IPv6-only address, no v4 fallback) on the client advertising `common.CapabilityIPv6` in `ConfigRequest.Capabilities` — see `Track.TargetsClient` in lantern-cloud's `cmd/api/track/track.go`. A client that can't actually reach the v6 internet would otherwise be handed an address it can never dial.
- `config/fetcher.go` now appends `C.CapabilityIPv6` to every `ConfigRequest` when `common.HasGlobalIPv6()` reports a real (non-ULA) global IPv6 route.
- `HasGlobalIPv6` and its interface-scanning helpers (`isGlobalIPv6`, `ifaceSnapshot`, `snapshotInterfaces`, `hasGlobalIPv6Using`) move from `vpn` to `common`, so both `vpn` (which already gates the TUN's own v6 ULA on it) and `config` can call it. `config` importing `vpn` directly hits a real import cycle — `vpn/boxoptions_test.go` is an internal test file that already imports `config`, and Go's test build rejects `vpn(test) → config → vpn`. Behavior and test coverage of the moved code are unchanged, just relocated to `common/ipv6.go` / `common/ipv6_test.go`.
- Bumps the `getlantern/common` pin to the commit that adds `CapabilityIPv6` (getlantern/common#37).

## Test plan
- [x] `go build ./config/... ./vpn/... ./common/... ./backend/...`
- [x] `go vet ./config/... ./vpn/... ./common/... ./backend/...` (no import cycle)
- [x] `go test ./common/... ./vpn/...` — all pass, including the relocated `TestIsGlobalIPv6` / `TestHasGlobalIPv6Using`
- [x] Confirmed `TestUserChangeSkipsConfigFetchWithoutCredentials` / `TestInitialAccountEventDoesNotLoopConfigCreation` failures in `./config/...` pre-exist on unmodified `main` (unrelated to this change)
- [ ] Once merged, the corresponding lantern-cloud change (`Track.TargetsClient` gate) can go out independently in either order — absence of the capability just excludes v6 tracks, which is today's behavior anyway

## Summary by CodeRabbit

* **New Features**
  * Added detection for active global IPv6 connectivity.
  * Configuration requests now advertise IPv6 capability when global IPv6 connectivity is available.
  * IPv6 tunnel address setup now uses the shared connectivity detection.

* **Bug Fixes**
  * Improved IPv6 detection across network interfaces, including cellular, VPN, and platform-specific configurations.
  * Excludes non-global addresses such as link-local, private, loopback, and IPv4-mapped addresses.


garmr-ulfr (changes requested):
As pointed out by both Copilot and CodeRabbit, `HasGlobalIPv6` doesn't verify internet reachability over IPv6. Additionally, we'd need to update the sing-box config, as it currently blocks all IPv6 traffic and forces IPv4 only:
https://github.com/getlantern/radiance/blob/main/vpn/boxoptions.go#L450
https://github.com/getlantern/radiance/blob/main/vpn/dnsoptions.go#L152

Do we need IPv6-only proxies? If a user were to switch from cellular with IPv6 to a WiFi network that's IPv4 only, they would no longer be able to reach those proxies. And they would lose all internet access until the next config fetch if they were only assigned IPv6-only proxies, which could take up to  10-15 minutes. I feel like we shouldn't have IPv6-only proxies.

AGMETEOR:
There's been some reports that IPv6 was performing well in some areas hence the decision to have some IPv6 proxies.

In case a user cannot connect through a proxy, don't they get switched over to another one by radiance?

AGMETEOR:
@garmr-ulfr why the decision to block IPv6?

garmr-ulfr:
> In case a user cannot connect through a proxy, don't they get switched over to another one by radiance?

They do through the auto-select outbound, I was saying if they're _only_ assigned IPv6-only proxies it can take a bit to cycle through all of them before it decides non are reachable and we need to fetch a new config. 

> @garmr-ulfr why the decision to block IPv6?

Oh, I'm not blocking enabling IPv6, the change request is because the tunnel would still block all IPv6 traffic due to the rules linked above. Those need to be changed too. 

Also, we'll need to enable IPv6 on all proxies or clients will fail to connect and just mark IPv4-only proxies as bad. If we want to mix IPv4-only and IPv6-only it's much more complicated.

garmr-ulfr:
> If we want to mix IPv4-only and IPv6-only it's much more complicated.

If we're able to just support IPv4 and IPv6 on all proxies, it would be significantly easier.

AGMETEOR:
> > If we want to mix IPv4-only and IPv6-only it's much more complicated.
> 
> If we're able to just support IPv4 and IPv6 on all proxies, it would be significantly easier.

So you're saying a proxy should have both a v6 and v4 address instead of either or. In that case would the client first dial v6 and then fallback to the v4 address?

garmr-ulfr:
> So you're saying a proxy should have both a v6 and v4 address instead of either or. In that case would the client first dial v6 and then fallback to the v4 address?

~Yes, exactly. Most applications prefer IPv6 over IPv4 and will try IPv6 first and fallback to IPv4 if it fails (happy-eyeballs). The caveat is applications received IPv6 DNS records and so tried to dial IPv6 over the IPv4-only proxies but don't fallback because they were told IPv6 should work. Which is the reason we had to add rules to block IPv6 altogether. If we enable both for all proxies, we would just need to remove the blocking rules and it should work.~

garmr-ulfr:
> > So you're saying a proxy should have both a v6 and v4 address instead of either or. In that case would the client first dial v6 and then fallback to the v4 address?
> 
> ~Yes, exactly. Most applications prefer IPv6 over IPv4 and will try IPv6 first and fallback to IPv4 if it fails (happy-eyeballs). The caveat is applications received IPv6 DNS records and so tried to dial IPv6 over the IPv4-only proxies but don't fallback because they were told IPv6 should work. Which is the reason we had to add rules to block IPv6 altogether. If we enable both for all proxies, we would just need to remove the blocking rules and it should work.~

Actually, after thinking about it more, this is at the application layer... well when we were having issues with IPv6 traffic, this kind of made sense from the logs and this explanation was only reinforced when adding the rules to block all IPv6 fixed it. But, the sing-box options only support one server address. I'll look into it more Monday. 

You can also test it out by removing those rules above and using the [CLI](https://github.com/getlantern/radiance/tree/main/cmd/lantern).

common/ipv6.go
```diff
@@ -0,0 +1,93 @@
...
+// unicast IPv6 address (not just ULA). Called per tunnel start; not cached.
+//
+// Two callers rely on this, for related but distinct reasons:
+//   - vpn gates whether the TUN gets its own IPv6 ULA — enabling it on a
+//     v4-only network has been observed to break things we haven't narrowed
+//     down.
+//   - config.fetcher gates whether to advertise common.CapabilityIPv6 to the
+//     server, which decides whether an IPv6-only proxy route (no v4
+//     fallback) is safe to assign.
+func HasGlobalIPv6() bool {
```
garmr-ulfr, line 62: As Copilot and CodeRabbit mentioned, this doesn't actually check if the internet can be reached over IPv6, just that a IPv6 public address was assigned. 

The comment is also inaccurate now as it's not only called on tunnel start. Would you change the comment to just
```go
// HasGlobalIPv6 reports whether the device has a non-loopback interface 
// carrying a global unicast IPv6 address (not just ULA). It does not verify 
// routing or internet connectivity.
func HasGlobalIPv6() bool {
```
so it's accurate. The rest of it is unnecessary and verbose.

config/fetcher.go
```diff
@@ -74,6 +74,18 @@ func (f *fetcher) fetchConfig(ctx context.Context, preferred common.PreferredLoc
...
 	}
+	// Advertise that we honor NonSelectableOutbounds (merge server-declared infra
+	// outbounds but keep them out of the proxy-selection groups) so the server can
+	// gate such outbounds on the capability rather than the client version.
+	capabilities := []string{C.CapabilityNonSelectableOutbounds}
+	// CapabilityIPv6 tells the server this device has a real route to the v6
+	// internet, not just a v6-capable OS, so it's safe to assign a proxy route
+	// whose address is IPv6-only (no v4 fallback). Reuses the same interface
+	// check vpn already trusts to decide whether the TUN itself gets a v6
+	// address, rather than a second, differently-tuned heuristic.
```
garmr-ulfr, line 85: Would you drop these comments? The second isn't accurate and they aren't needed anyway.

garmr-ulfr (approved):
Ignore everything I said, this is fine and everything should work. Just a couple of comment changes/cleanup, but LGTM otherwise.

